PR #2, opened 2024-03-15
Additions

harshilsharma63:
This PR is not meant to be merged.

The PR serves the purpose of a full plugin code review.

#### Summary
Changes made-
1. Renamed plugin from `mattermost-starter-template` to `mattermost-plugin-user-survey`.
2. Upgraded golangci-lint from 1.51 to 1.56.
3. Upgraded server and webapp dependencies
4. Updated plugin manifest
5. Added SQL migrations
6. Added survey date and time selector system console settings


harshilsharma63:
The CI doesn't work because I'm using methods in server/public package that are not yet published. Will fix it later.

build/pluginctl/logs.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	logsPerPage     = 100 // logsPerPage is the number of log entries to fetch per API call
+	logsPerPage     = 100 // logsPerPage is the number of log entries to fetch per PluginAPI call
```
harshilsharma63, line 18: Note to reviewer - these `API` to `PluginAPI` text changes were accidental and are already fixed in `main`.

webapp/webpack.config.js.disabled
```diff
@@ -0,0 +1,103 @@
+const exec = require('child_process').exec;
```
harshilsharma63, line 1: Note to reviewer - This file isn't needed and is already deleted on `main`.

webapp/src/components/common/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,43 @@
...
+    value?: DropdownOption
+    defaultValue?: DropdownOption
+    onChange: (newValue: DropdownOption) => void
+}
+const Dropdown = ({options, value, defaultValue, onChange}: Props) => {
+    const customComponents = useMemo(() => {
+        return {
+            Control,
+        };
+    }, []);
```
larkox, line 25: This can be simplified to something like:

`const customComponents = useMemo(() => ({Control}), []);`

webapp/src/components/common/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    // This handler only serves the purpose of satisfying typescript.
+    // Otherwise it complains abut incorrect type's callback passed to Select's onChange prop.
```
larkox, line 28: We control the type of onChange, right? Can nothing be done on that side so this is more elegant?

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
```diff
@@ -0,0 +1,40 @@
...
+export type Props = {
+    value: Date
+    onChange?: (value: Date) => void
+};
+
+const SurveyDateSelector = ({value, onChange}: Props) => {
+    const onSelectHandler = useCallback((newValue: Date) => {
+        if (onChange) {
+            onChange(newValue);
+        }
```
larkox, line 20: This can be simplified to `onChange?.(newValue);`

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useCallback} from 'react';
+import {format} from 'date-fns';
```
larkox, line 5: I wonder if we can/should use intl to format the date. 0/5

webapp/src/components/surveyTimeSelector/surveyTimeSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }, []);
+
+    const onChangeHandler = useCallback((newValue: DropdownOption) => {
+        onChange(newValue.value);
+    }, [onChange]);
+
+    return (
+        <div className='SurveyTimeSelector'>
+            <Dropdown
+                value={value ? {value, label: value} : options[18]}
```
larkox, line 42: New object on every render. Also, not a big fan of the magic number, but probably is fine in this context.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -0,0 +1,81 @@
...
+        const dateTimeConfig = config.PluginSettings?.Plugins?.['com.mattermost.user-survey']?.surveydatetime;
+
+        if (dateTimeConfig?.time) {
+            setSurveyTime(dateTimeConfig.time);
+        }
+
+        if (dateTimeConfig?.date) {
+            setSurveyDate(parse(dateTimeConfig.date, 'dd/MM/yyyy', new Date()));
+        }
+    }, [config.PluginSettings, config.PluginSettings?.Plugins]);
```
larkox, line 34: I think we can improve this dependency as just:
`config.PluginSettings?.Plugins?.['com.mattermost.user-survey']?.surveydatetime`
It is more specific of what we need, so it renders less often. We could argue that this won't change much eitherway in this context and have something this explicit is bad for readability.

In any case, having these two dependencies seems unnecessary. One of them should suffice.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const [surveyTime, setSurveyTime] = useState<string>(defaultSurveyTime);
+    const [surveyDate, setSurveyDate] = useState<Date>(defaultSurveyDate);
+
+    useEffect(() => {
+        // sets the date picker and time picker to the values saved in config on load
+
+        const dateTimeConfig = config.PluginSettings?.Plugins?.['com.mattermost.user-survey']?.surveydatetime;
+
+        if (dateTimeConfig?.time) {
+            setSurveyTime(dateTimeConfig.time);
```
larkox, line 28: Why do we set a state in an effect (creating an extra render) instead of using the value from the props directly on render?

If it is about reducing the amount of times we parse the time string, we can use a useMemo instead.

larkox (approved):
A few nitpicks here and there, but nothing blocking. Great work!

.nvmrc
```diff
@@ -1 +1 @@
-16.13.1
```
devinbinnie, line 1: Unrelated but we should probably update the template here too, OOS of this PR of course :P

server/store/migrations/000001_create_survey_table.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS {{.prefix}}survey (
```
devinbinnie, line 1: Should we maybe label these as SQL templates and not `.sql` files? They do require being run through the template engine before they can run so maybe we should call them `.sqltemp` files or something like that?

server/store/sqlstore.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/url"
+
+	"github.com/mattermost/mattermost/server/public/plugin"
+	"github.com/mattermost/mattermost/server/public/pluginapi/cluster"
+
+	"github.com/mattermost/mattermost-plugin-user-survey/server/model"
+)
+
+const (
+	TablePrefix = "user_survey_"
```
devinbinnie, line 14: Just curious - why is this configurable? Will we find ourselves needing to change this at some point? Or is it so that people can whitelabel the plugin?

server/model/database.go
```diff
@@ -0,0 +1,6 @@
+package model
+
+const (
+	DBTypeMySQL    = "mysql"
```
devinbinnie, line 4: We really should have these constants in the `public/model` package, I was surprised to find out we didn't lol.
Don't need to change this now of course :P

webapp/src/components/common/datePicker/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+    closeOnSelect?: boolean
+};
+
+const DatePicker = ({value, children, onSelect, closeOnSelect}: Props) => {
+    const [isPopupOpen, setPopupOpen] = useState<boolean>(false);
+
+    const onSelectHandler = useCallback((day: Date) => {
+        onSelect(day);
+
+        // Checking it this way to handle null and undefined.
```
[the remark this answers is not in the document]
devinbinnie: We generally have a rule against checking for negative cases, it's actually preferred to check for positive so I'd agree with that change.

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
devinbinnie:
```suggestion
        onChange?.(newValue);
```
Always good to make it easy to change ;)

webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -1,5 +1,35 @@
...
+
+export type Config = {
+    PluginSettings?: {
+        Plugins?: {
+            'com.mattermost.user-survey'?: {
+                surveydatetime: DateTimeConfig,
+            },
+        },
+    },
+    [key: string]: unknown,
```
devinbinnie, line 18: Can we either import this type from `mattermost-redux` or get rid of this arbitrary dictionary? I always find they just kinda muddle/defeat the purpose of typing, unless we explicitly expect to find any potential value in there.

devinbinnie (approved):
Looking good! Just a few random comments/suggestions/laments about other pieces of code :P

plugin.json
```diff
@@ -1,9 +1,9 @@
 {
-    "id": "com.mattermost.plugin-starter-template",
-    "name": "Plugin Starter Template",
-    "description": "This plugin serves as a starting point for writing a Mattermost plugin.",
-    "homepage_url": "https://github.com/mattermost/mattermost-plugin-starter-template",
-    "support_url": "https://github.com/mattermost/mattermost-plugin-starter-template/issues",
+    "id": "com.mattermost.user-survey",
+    "name": "User Survey",
+    "description": "A plugin designed for gathering user feedback on Mattermost and related topics.",
```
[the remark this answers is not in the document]
harshilsharma63: Created a Jira ticket for text review.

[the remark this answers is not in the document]
harshilsharma63: https://mattermost.atlassian.net/browse/MM-57313

webapp/src/components/common/datePicker/index.tsx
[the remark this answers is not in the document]
harshilsharma63: I've updated this to `(closeOnSelect || typeof closeOnSelect === 'undefined')`. If the prop isn't specified, it will be undefined.

webapp/src/components/common/dropdown/dropdown.tsx
harshilsharma63: Done 👍

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
hmhealey: Note that doing that might require naming it before it gets to this point. I feel like ESLint would complain that the current code doesn't depend on `config`

webapp/src/types/mattermost-webapp/index.d.ts
hmhealey: It could extend `GlobalState` from `@mattermost/types`. That only contains types for `state.entities`, but that's most of what you'd need anyway

webapp/.eslintrc.json
hmhealey, line 1: Not necessary now since you've already gone through all this work, but you could instead have this use https://www.npmjs.com/package/@mattermost/eslint-plugin now that that's finally published. If there's no overrides needed for this repo, it should be as easy as adding that package and then replacing all this with
```json
{
  "root": true,
  "extends": [
    "plugin:@mattermost/react"
  ]
}
```

webapp/package.json
```diff
@@ -5,86 +5,66 @@
...
-    "eslint-plugin-cypress": "2.12.1",
+    "eslint-import-resolver-typescript": "3.6.1",
+    "eslint-import-resolver-webpack": "0.13.8",
+    "eslint-plugin-cypress": "2.15.1",
     "eslint-plugin-header": "3.1.1",
-    "eslint-plugin-import": "2.25.4",
-    "eslint-plugin-jest": "26.5.3",
+    "eslint-plugin-import": "2.29.1",
+    "eslint-plugin-jest": "27.9.0",
     "eslint-plugin-mattermost": "github:mattermost/eslint-plugin-mattermost",
```
hmhealey, line 36: Also, that'd mean you can get rid of the old ESLint plugin which is a git dependency and causes NPM to complain because of a non-existent vulnerability in it

webapp/src/components/common/datePicker/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+        // The default behavior is to close on select.
+        if (closeOnSelect !== false) {
+            setPopupOpen(false);
+        }
+    }, [closeOnSelect, onSelect]);
+
+    const {context} = useFloating({
```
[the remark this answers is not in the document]
hmhealey: Given the examples on Floating UI's site don't use it, I don't think it's needed here

webapp/src/components/common/datePicker/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+        useDismiss(context, {
+            enabled: true,
+            outsidePress: true,
+        }),
+    ]);
+
+    const popperRef = useRef<HTMLDivElement>(null);
+
+    return (
+        <div className='DatePicker'>
```
hmhealey, line 62: Have you double checked to make sure that this doesn't conflict with any classes defined by the web app? I can't find any obvious conflict at least, but I wanted to mention it

webapp/src/components/common/datePicker/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+                        <div
+                            className='rdp_wrapper'
+                            {...getFloatingProps()}
+                        >
+                            <DayPicker
+                                selected={value}
+                                defaultMonth={value}
+                                className='DatePicker-day-picker'
+                                disabled={{
+                                    before: new Date(),
```
hmhealey, line 87: This should probably be saved in a ref since it's value will change every time this is rendered

webapp/src/components/common/datePicker/style.scss
hmhealey, line 1: Nit: The web app uses 4 spaces for indenting CSS

webapp/src/components/common/datePicker/style.scss
```diff
@@ -0,0 +1,60 @@
...
+    border-radius: 4px;
+    box-shadow: 0 8px 24px 0 rgba(0, 0, 0, 0.12);
+  }
+}
+
+.DatePicker {
+  position: relative;
+}
+
+.rdp-day {
```
hmhealey, line 22: Is there any risk of these types conflicting with types specified by the web app? I'd assume that they would be the same, but I know that this is in the System Console where those types might not be applied

hmhealey: Actually, if these are in the System Console, how does this look on a dark theme? Do these variables even work there?

webapp/src/components/common/dropdown/dropdown.tsx
hmhealey: If this should handle selecting multiple objects, I'd suggest making the this component's `onChange` always expect an array so that the caller can consistently handle that instead of having to figure out the type. This component could then handle wrapping a single result in an array

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
hmhealey: I'll go a step further actually. Couldn't we just pass `onChange` into the `DatePicker`'s `onSelect` prop? It doesn't look like we need to declare a new handler here

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
```diff
@@ -0,0 +1,40 @@
...
+            onChange(newValue);
+        }
+    }, [onChange]);
+
+    return (
+        <DatePicker
+            onSelect={onSelectHandler}
+            value={value}
+        >
+            <div className='SurveyDateSelector form-control'>
```
hmhealey, line 28: Not sure if it's something you can or want to change, but `form-control` is actually a CSS class from Bootstrap that's leaking into here. I'd personally try to get rid of it, but if we need it for consistent styling, then it's totally fine to leave it

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -0,0 +1,81 @@
...
+import SurveyTimeSelector from 'components/surveyTimeSelector/surveyTimeSelector';
+import SurveyDateSelector from 'components/surveyDateSelector/surveyDateSelector';
+import {CustomComponentProps} from 'types/mattermost-webapp';
+
+import './style.scss';
+
+const defaultSurveyTime = '09:00';
+
+// date 30 days from today
+const defaultSurveyDate = new Date(new Date().setDate((new Date()).getDate() + 30));
```
hmhealey, line 16: This is 30 days from the moment this file is loaded which seems like it might not be quite what you want. It may be better to save it in a ref in the component just in case someone leaves MM open for a few days.

Also, the JS `Date` type is awful to use, but I think you might be able to simplify this a bit to get rid of so many date constructors and brackets if you break it into two lines
```js
const defaultSurveyDate = new Date();
defaultSurveyDate.setDate(defaultSurveyDate.getDate() + 30);
```

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -0,0 +1,81 @@
...
+        onChange(id, setting);
+    }, [id, onChange, setSaveNeeded, surveyDate, surveyTime]);
+
+    const surveyTimeChangeHandler = useCallback((value: string) => {
+        setSurveyTime(value);
+    }, []);
+
+    const surveyDateChangeHandler = useCallback((value: Date) => {
+        setSurveyDate(value);
+    }, []);
```
hmhealey, line 57: These seem like they could also be passed directly into the children

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    return (
+        <div className='SurveyDateTime'>
+            <div className='row'>
```
hmhealey, line 61: `row` is another Bootstrap class if you case

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -0,0 +1,81 @@
...
+                />
+                <SurveyDateSelector
+                    value={surveyDate}
+                    onChange={surveyDateChangeHandler}
+                />
+            </div>
+
+            <div className='row'>
+                <p>
+                    {`A bot message with the survey will be sent to all users at ${surveyTime} UTC on ${format(surveyDate, 'do MMM yyyy')}.`}
```
hmhealey, line 74: I'm not sure if we have a good way yet to make plugins translatable, but I always like to mention it

hmhealey (changes requested):

webapp/src/components/common/datePicker/index.tsx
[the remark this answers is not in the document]
harshilsharma63: Yeah, its not needed.

webapp/src/components/common/dropdown/dropdown.tsx
harshilsharma63: our onChange is very simple and its abstracted from the fact that this component uses react-select underneath it. Its the react-select's onChange thats complicated and I don't want to pass that complication to users of our dropdown component, hence this wrapper.

harshilsharma63: @hmhealey I'll do that when I add the multiselector for team include/exclude setting but yeah, it will become an array.

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
harshilsharma63: Done @hmhealey 's suggestion. There was nothing else happening in the extra change handler to just passed prop change handler directly to the component. #

webapp/src/components/surveyTimeSelector/surveyTimeSelector.tsx
harshilsharma63: Fixed.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: Having just `config.PluginSettings?.Plugins` was sufficient. And yeah, if I set it to `config.PluginSettings?.Plugins?.['com.mattermost.user-survey']?.surveydatetime` ESlint complains of having too complex expression in dependency array. So I think we can leave it to just `config.PluginSettings?.Plugins`.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: Over here we're setting the value from props into the state. A value may not always exist in the props, for example in case of first time installing the plugin. So we need to fallback to a default value (`defaultSurveyTime`). So I start with a state populated by the default value, then use useEffect to check if a value is present in props. The logic is lengthy enough to not be put in useState(), thats why.

.nvmrc
harshilsharma63: The plugin starter template has much more to update, like getting rid of babel, going full typescript, updating nodejs etc.. I don't think we have the resources right now for it but when time comes.

server/store/migrations/000001_create_survey_table.up.sql
harshilsharma63: We always call them .sql in focalboard and playbooks, so just following convention.

server/store/sqlstore.go
harshilsharma63: We'll use a different prefix in tests, for example. But other than that, its just here instead of being hardcoded, kinda like a central place.

server/model/database.go
harshilsharma63: Agreed.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
larkox: ESLint will warn, but will not fail. But good practice to store the "actual value you need" in a variable so it is more readable.

webapp/src/types/mattermost-webapp/index.d.ts
harshilsharma63: Checking the value received in here, its `AdminConfig`, so set it to that.

webapp/.eslintrc.json
harshilsharma63: Done!

webapp/src/components/common/datePicker/index.tsx
harshilsharma63: yes, its not used in MM webapp.

webapp/src/components/common/datePicker/index.tsx
harshilsharma63: Makes sense. Done.

webapp/src/components/common/datePicker/style.scss
harshilsharma63: I went ahead and added stylelint to the plugin with the same config as MM webapp. Also updated package.json lint and fix script names.

harshilsharma63: This fixes the 4 space issue and ensures it stays fixed.

webapp/src/components/common/datePicker/style.scss
harshilsharma63: So these react day picker (rpd) styles are actually copy-pasted from MM webapp and only the needed parts kept, so for now, they are duplicate but they don't conflict. 

System console doesn't have a dark theme so it always uses the same colors.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: I've added a Jira ticket for adding support for this. Its possible (Playbook does it) but it increases the client bundle size exponentially. Will investigate and add it later.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: Renamed it to `horizontal`, the more classname conflicts we can avoid the better.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: Done.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: Fixed. Set the state to 30 days from today if a saved config is not found.

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
harshilsharma63: yeah it's for styling. It applies a loooooot of styling that was taking too long to apply manually so I just used the class name.

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
harshilsharma63: Done.

harshilsharma63:
@devinbinnie @larkox @hmhealey since this PR was between two orphan branches, I've fixed all issues in a PR into `main` here - #4 .

harshilsharma63:
Closing as #4 is merged.

PR #3, opened 2024-03-15
Prevented save button from being activated on load even when nothing changes

harshilsharma63:
#### Summary
Currently, the Save button in settings gets activated on load. This was because I was calling the "save needed" function in useEffect, which gets called on load as well. Converted it to a useCallback function to resolve it.


M-ZubairAhmed (approved):

esarafianou:
@harshilsharma63 Had a peek, no need for my review, all good

hmhealey (approved):

PR #4, opened 2024-03-21
Full Code Review Fixes

harshilsharma63:
#### Summary
1. Fixes for this PR - https://github.com/mattermost/mattermost-plugin-user-survey/pull/2
2. Fixed server and webapp tests
3. Removed pluginctl/logs as its not needed.
4. Upgraded MM `server/public` version to v0.1.
5. Added Stylelint
6. Used MM eslint plugin.


webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
```diff
@@ -1,36 +1,41 @@
...
-    const onSelectHandler = useCallback((newValue: Date) => {
-        if (onChange) {
-            onChange(newValue);
-        }
-    }, [onChange]);
+    const [formatedDate, setFormattedDate] = useState<string>('');
+
+    useEffect(() => {
+        setFormattedDate(value.toLocaleDateString(undefined, dateFormattingOptions));
+    }, [value]);
```
larkox, line 26: This causes an extra re-render every time the prop changes. Unless there are reasons to perform that extra re-render, you can change this to a `useMemo`.

Also, not sure how expensive is `toLocalDateString`, but if it is cheap, we could remove the useMemo also.

webapp/src/components/surveyTimeSelector/surveyTimeSelector.tsx
```diff
@@ -32,14 +33,21 @@ const SurveyTimeSelector = ({value, onChange}: Props) => {
         return timeStrings;
     }, []);
 
+    const [dropdownValue, setDropdownValue] = useState<DropdownOption>(options[18]);
+    useEffect(() => {
+        if (value) {
+            setDropdownValue({value, label: value});
+        }
+    }, [value]);
```
larkox, line 41: Similar comment about my previous one. You can use `useMemo` here instead of using a state.

In this case we don't want to remove the `useMemo` because if not, we would be passing a new object to the prop on every render.

larkox (approved):
Just a question and a couple of nitpicks.

webapp/src/components/common/datePicker/index.tsx
```diff
@@ -14,25 +11,28 @@ import {
...
     const [isPopupOpen, setPopupOpen] = useState<boolean>(false);
+    const currentDate = useRef<Date>(new Date());
 
     const onSelectHandler = useCallback((day: Date) => {
         onSelect(day);
 
         // Checking it this way to handle null and undefined.
         // The default behavior is to close on select.
-        if (closeOnSelect !== false) {
+        if (closeOnSelect || typeof closeOnSelect === 'undefined') {
```
devinbinnie, line 35: Just a thought - as a simplification can we rework this to work off a positive case instead, so you could instead write something like `if (dontCloseOnSelect)`

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
```diff
@@ -1,36 +1,41 @@
...
+    onChange: (value: Date) => void;
 };
 
 const SurveyDateSelector = ({value, onChange}: Props) => {
-    const onSelectHandler = useCallback((newValue: Date) => {
-        if (onChange) {
-            onChange(newValue);
-        }
-    }, [onChange]);
+    const [formatedDate, setFormattedDate] = useState<string>('');
```
devinbinnie, line 22: Nit: Don't think you need the generic here if you're specifying a default value.

devinbinnie (approved):
Couple more non-blocking comments, otherwise LGTM.

webapp/src/components/surveyTimeSelector/surveyTimeSelector.tsx
hmhealey: Yeah, I agree with this to get rid of the extra render here. I was a bit confused as to why we had value as both a prop and as state when reading this before

hmhealey (approved):
Other than the one thing I left feedback on, this looks good to me

webapp/.eslintignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 webpack.config.ts
 dist
+manifest.ts
```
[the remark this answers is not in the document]
harshilsharma63: Because its an auto-generated file. Any changes we make in there are always reverted.

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
harshilsharma63: yeah, `useMemo` makes more sense.

harshilsharma63: Done.

webapp/src/components/surveyTimeSelector/surveyTimeSelector.tsx
harshilsharma63: Done.

webapp/src/components/common/datePicker/index.tsx
harshilsharma63: Fixed by adding a default value where I'm expanding the props. So I don't get null and undefined states here, only true or false. So this is now simplified to just `if (dontCloseOnSelect)`.

webapp/src/components/surveyDateSelector/surveyDateSelector.tsx
harshilsharma63: Yeah but I just specify it anyways everywhere out of habit :P

harshilsharma63:
Thanks a ton @hmhealey @danielsischy @devinbinnie for the review~

PR #5, opened 2024-03-28
Team selector and survey expiry setting

harshilsharma63:
#### Summary
1. Added setting to specify how long the survey will last.
1.  Added setting to filter out teams from survey

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-57459

### DIFF
--- a/webapp/src/components/common/multiSelect/multiselect.tsx
+++ b/webapp/src/components/common/multiSelect/multiselect.tsx
@@ -1,0 +1,38 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useCallback, useMemo, useState} from 'react';
+
+import './style.scss';
+import {DropdownOption} from 'components/common/dropdown/dropdown';
+import Select, {type MultiValue, type SingleValue} from 'react-select';
+
+export type Props = {
+    options: DropdownOption[];
+    values?: DropdownOption[];
+    customComponents?: any;
+}
+
+function Multiselect({options, values, customComponents}: Props) {
+    const [val, setVal] = useState<DropdownOption[]>([]);
+
+    const onChangeHandler = useCallback((newValue: SingleValue<DropdownOption> | MultiValue<DropdownOption>) => {
+        if (!Array.isArray(newValue)) {
+            setVal(existing => {
+                existing.push(newValue);
+                return existing;
+            });
+        }
+    }, []);
+
+    return (
+        <Select
+            value={val}
+            options={options}
+            components={customComponents}
+            onChange={onChangeHandler}
+        />
+    );
+}
+
+export default Multiselect;
--- a/webapp/src/components/common/multiSelect/multiselect.tsx
+++ b/webapp/src/components/common/multiSelect/multiselect.tsx
@@ -1,32 +1,34 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useCallback, useMemo, useState} from 'react';
+import React, {useCallback} from 'react';
 
 import './style.scss';
 import {DropdownOption} from 'components/common/dropdown/dropdown';
-import Select, {type MultiValue, type SingleValue} from 'react-select';
+import Select, {GroupBase, type MultiValue, type SingleValue} from 'react-select';
+import {SelectComponentsConfig} from 'react-select/dist/declarations/src/components';
+
+export type CustomComponentsDefinition = SelectComponentsConfig<DropdownOption, true, GroupBase<DropdownOption>>
 
 export type Props = {
     options: DropdownOption[];
     values?: DropdownOption[];
-    customComponents?: any;
+    customComponents?: CustomComponentsDefinition;
+    onChange: (selectedValues: DropdownOption[]) => void;
 }
 
-function Multiselect({options, values, customComponents}: Props) {
-    const [val, setVal] = useState<DropdownOption[]>([]);
-
+function Multiselect({options, values, customComponents, onChange}: Props) {
     const onChangeHandler = useCallback((newValue: SingleValue<DropdownOption> | MultiValue<DropdownOption>) => {
         if (Array.isArray(newValue)) {
-            setVal(newValue);
+            onChange(newValue);
         }
-    }, []);
+    }, [onChange]);
 
     return (
         <Select
             isMulti={true}
             isClearable={false}
-            value={val}
+            value={values}
             options={options}
             components={customComponents}
             onChange={onChangeHandler}
--- a/webapp/src/components/common/multiSelect/multiselect.tsx
+++ b/webapp/src/components/common/multiSelect/multiselect.tsx
@@ -17,16 +17,15 @@ function Multiselect({options, values, customComponents}: Props) {
     const [val, setVal] = useState<DropdownOption[]>([]);
 
     const onChangeHandler = useCallback((newValue: SingleValue<DropdownOption> | MultiValue<DropdownOption>) => {
-        if (!Array.isArray(newValue)) {
-            setVal(existing => {
-                existing.push(newValue);
-                return existing;
-            });
+        if (Array.isArray(newValue)) {
+            setVal(newValue);
         }
     }, []);
 
     return (
         <Select
+            isMulti={true}
+            isClearable={false}
             value={val}
             options={options}
             components={customComponents}
--- a/webapp/src/components/common/multiSelect/multiselect.tsx
+++ b/webapp/src/components/common/multiSelect/multiselect.tsx
@@ -2,12 +2,11 @@
 // See LICENSE.txt for license information.
 
 import React, {useCallback} from 'react';
+import type {GroupBase, MultiValue, SingleValue} from 'react-select';
+import Select from 'react-select';
+import type {SelectComponentsConfig} from 'react-select/dist/declarations/src/components';
 
-import './style.scss';
-import {DropdownOption} from 'components/common/dropdown/dropdown';
-import Select, {GroupBase, type MultiValue, type SingleValue} from 'react-select';
-import {SelectComponentsConfig} from 'react-select/dist/declarations/src/components';
-
+import type {DropdownOption} from 'components/common/dropdown/dropdown';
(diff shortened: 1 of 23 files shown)


esarafianou:
@harshilsharma63 had a quick look, no need for any security review.

webapp/src/components/systemConsole/teamFilter/teamIcon/teamIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+    // @ts-expect-error last_team_icon_update
+    // field will be included in latest types release,
+    // but its there in the data already
+    if (data.raw.last_team_icon_update) {
+        // last_team_icon_update is set whenever a team icon is set or updated.
+        // So, a truthy value of `last_team_icon_update` field indicates precense of
+        // team icons, so here we use it to show the team icon image.
+
+        // @ts-expect-error sane reason as above about the last_team_icon_update field
+        const teamURL = Client4.getTeamIconUrl(data.value, (data.raw as Team)?.last_team_icon_update);
```
larkox, line 27: We can omit most of this if we just check for `last_team_icon_update` to be in the raw. Something like this:

```suggestion
    if (`last_team_icon_update` in data.raw && typeof(data.raw.last_team_icon_update) === 'number' && data.raw.last_team_icon_update) {
        const teamURL = Client4.getTeamIconUrl(data.value, data.raw.last_team_icon_update);
```

larkox (approved):
Looks great.

Just one real suggestion, and a couple considerations from my side.

webapp/src/components/common/multiSelect/multiselect.tsx
```diff
@@ -0,0 +1,38 @@
...
+    values?: DropdownOption[];
+    customComponents?: CustomComponentsDefinition;
+    onChange: (selectedValues: DropdownOption[]) => void;
+}
+
+function Multiselect({options, values, customComponents, onChange}: Props) {
+    const onChangeHandler = useCallback((newValue: SingleValue<DropdownOption> | MultiValue<DropdownOption>) => {
+        if (Array.isArray(newValue)) {
+            onChange(newValue);
+        }
```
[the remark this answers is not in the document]
harshilsharma63: This is a multiselect component, so the `newValue` will always be an array. However the underlying react-select component supports both single and multi select, so  its change handler wither receives a `DropdownOption`, or an array of `DropdownOption`. Since I know this is a multiselect component, I know it will always be an array, so I check for array as a way to satisfy type checks. The change handler of this custom component always expects an array, so either I can force `newValue as DropdownOption[]`, or check it this way.

webapp/src/components/systemConsole/teamFilter/teamFilter.tsx
```diff
@@ -0,0 +1,96 @@
...
+        };
+
+        task();
+    }, [config.PluginSettings.Plugins]);
+
+    const customComponents: CustomComponentsDefinition = useMemo(() => (
+        {
+            Option: CustomOption,
+            MultiValueContainer: CustomMultiValueContainer,
+
```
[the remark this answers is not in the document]
harshilsharma63: Fixed.

webapp/src/components/systemConsole/teamFilter/teamFilter.tsx
```diff
@@ -0,0 +1,96 @@
...
+import type {CustomComponentProps, TeamFilterConfig} from 'types/mattermost-webapp';
+
+function TeamFilter({id, setSaveNeeded, onChange, config}: CustomComponentProps) {
+    const [selectedTeams, setSelectedTeams] = useState<DropdownOption[]>([]);
+    const [allTeamsOptions, setAllTeamsOptions] = useState<DropdownOption[]>([]);
+
+    // fetch all teams to populate options
+    useEffect(() => {
+        const task = async () => {
+            const teams: Team[] = await Client4.getTeams(0, 10000, false) as Team[];
```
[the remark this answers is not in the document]
harshilsharma63: Thanks for starting up a conversation 👍

webapp/src/components/systemConsole/teamFilter/teamFilter.tsx
```diff
@@ -0,0 +1,96 @@
...
+                    const team = options.find((option) => option.value === teamId);
+                    return {
+                        label: team?.label || `Archived Team: ${teamId}`,
+                        value: teamId,
+                        raw: team?.raw,
+                    };
+                });
+
+                setSelectedTeams(intialOptions);
+            }
```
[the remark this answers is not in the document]
harshilsharma63: Fixed by keeping a map of team id -> team for this purpose.

webapp/src/components/systemConsole/teamFilter/teamIcon/teamIcon.tsx
harshilsharma63: I updated the `@mattermost/types` package to bring in the `last_team_icon_update` field, so no more need of all this checks. The field now exists in type definition.

webapp/src/components/systemConsole/teamFilter/teamIcon/teamIcon.tsx
```diff
@@ -15,16 +15,9 @@ export type Props = {
...
-    // field will be included in latest types release,
-    // but its there in the data already
-    if (data.raw.last_team_icon_update) {
-        // last_team_icon_update is set whenever a team icon is set or updated.
-        // So, a truthy value of `last_team_icon_update` field indicates precense of
-        // team icons, so here we use it to show the team icon image.
-
-        // @ts-expect-error sane reason as above about the last_team_icon_update field
-        const teamURL = Client4.getTeamIconUrl(data.value, (data.raw as Team)?.last_team_icon_update);
+    const teamIconLastUpdated = (data.raw as Team).last_team_icon_update;
```
larkox, line 18: We can leave it like this (it is good enough and we control the data, so probably this is way more readable that what I am going to propose), but type casting is quite insecure in Typescript (unknown can be cast to anything) and we should try to avoid it.

In order to make sure it is a team, the "Typescript way" is to use a guard. Another way to handle this, is to check if `last_team_icon_update` exist in raw.

Nevertheless, as I said, let's keep it like this, since it is more readable and has less boilerplate.

larkox (approved):
LGTM

webapp/src/components/systemConsole/expiry/expiry.tsx
```diff
@@ -0,0 +1,71 @@
...
+// See LICENSE.txt for license information.
+
+import type {ChangeEvent} from 'react';
+import React, {useCallback, useEffect, useState} from 'react';
+
+import type {CustomComponentProps, ExpiryConfig} from 'types/mattermost-webapp';
+
+import './style.scss';
+
+const defaultExpiry = '30';
```
M-ZubairAhmed, line 11: shouldnt constants be UPPER_CASED?

webapp/src/components/systemConsole/expiry/expiry.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const Expiry = ({id, setSaveNeeded, onChange, config}: CustomComponentProps) => {
+    const [expiryDays, setExpiryDays] = useState<string>(defaultExpiry);
+    const [error, setError] = useState<string>('');
+
+    useEffect(() => {
+        // Set initial value from saved config
+
+        const expiryConfig = config.PluginSettings?.Plugins?.['com.mattermost.user-survey']?.surveyexpiry;
```
M-ZubairAhmed, line 20: `com.mattermost.user-survey` do you want to extract this as well as a constant

webapp/src/components/systemConsole/expiry/expiry.tsx
```diff
@@ -0,0 +1,71 @@
...
+            <div className='vertical'>
+                {
+                    error &&
+                    <p className='errorMessage'>
+                        {error}
+                    </p>
+                }
+                <p>
+                    {'Select the number of days for which the survey will stay active and users will be able to respond to it. No more responses for this survey will be accepted after the set number of days have passed. '}
+                </p>
```
M-ZubairAhmed, line 65: does plugins not have i18n?

webapp/src/components/systemConsole/teamFilter/teamFilter.tsx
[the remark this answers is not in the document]
M-ZubairAhmed: No this is not how its implementation is updated take a look here
https://github.com/mattermost/mattermost/blob/978f335925b91a3e92c87766caf4f4a638d0760b/webapp/channels/src/components/admin_console/filter/team_filter_dropdown/team_filter_dropdown.tsx

M-ZubairAhmed (approved):
Approving, some questions

webapp/src/components/systemConsole/teamFilter/teamFilter.tsx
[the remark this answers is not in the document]
harshilsharma63: We can do the same later if either need be or we have time.

webapp/src/components/systemConsole/expiry/expiry.tsx
harshilsharma63: Its fixed in #6 , which is a PR on top of this PR. All such constants are upper cased there.

webapp/src/components/systemConsole/expiry/expiry.tsx
larkox: Maybe we can use `manifest.id` for this one?

webapp/src/components/systemConsole/expiry/expiry.tsx
larkox: Not really. IIRC we don't have a good way to send this to weblate, and not sure we have a good way to merge with the intl provider of the webapp.

@mickmister May know better what is the current "state of the art" on this topic.

webapp/src/components/systemConsole/expiry/expiry.tsx
harshilsharma63: The key is part of type definition in `webapp/src/types/mattermost-webapp/index.d.ts` so can't be fudged without changing the type declaration.

webapp/src/components/systemConsole/expiry/expiry.tsx
harshilsharma63: They do but I believe the webapp bundle then includes all language files and increases bundle size a lot. I have a ticket in jira for managing this later on when needed.

PR #6, opened 2024-04-04
Survey question section

harshilsharma63:
#### Summary
1. Added section to customize question text
2. Created a single custom config component and moved all existing custom configs into it. This had to be done to resolve a few cases which were otherwise impossible to do without getting redux involved in the plugin. Even this, this is the simplest and fastest way to solve a lot of problems.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-56692

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -53,6 +53,7 @@
     "classnames": "2.5.1",
     "core-js": "3.36.0",
     "date-fns": "3.3.1",
+    "debounce-promise": "3.1.2",
     "enzyme": "3.11.0",
     "enzyme-to-json": "3.6.2",
     "mattermost-redux": "5.33.1",
@@ -61,7 +62,8 @@
     "react-redux": "8.0.5",
     "react-select": "5.8.0",
     "redux": "4.2.1",
-    "typescript": "4.9.5"
+    "typescript": "4.9.5",
+    "use-debounce": "10.0.0"
   },
   "jest": {
     "preset": "ts-jest",
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -53,7 +53,6 @@
     "classnames": "2.5.1",
     "core-js": "3.36.0",
     "date-fns": "3.3.1",
-    "debounce-promise": "3.1.2",
     "enzyme": "3.11.0",
     "enzyme-to-json": "3.6.2",
     "mattermost-redux": "5.33.1",
--- a/webapp/src/components/systemConsole/enableSurvey/index.tsx
+++ b/webapp/src/components/systemConsole/enableSurvey/index.tsx
@@ -7,18 +7,18 @@ import type {CustomSettingChildComponentProp} from 'components/systemConsole/ind
 
 import './style.scss';
 
-function EnableSurvey({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) {
-    const defaultValue = false;
+const DEFAULT_ENABLE_SURVEY = false;
 
-    const [enabled, setEnabled] = useState<boolean>(defaultValue);
+function EnableSurvey({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) {
+    const [enabled, setEnabled] = useState<boolean>(DEFAULT_ENABLE_SURVEY);
 
     useEffect(() => {
         const enabledSurveyConfig = config.PluginSettings.Plugins['com.mattermost.user-survey']?.systemconsolesetting.EnableSurvey;
 
-        const initialValue = enabledSurveyConfig || defaultValue;
+        const initialValue = enabledSurveyConfig || DEFAULT_ENABLE_SURVEY;
         setEnabled(initialValue);
         setInitialSetting(id, initialValue);
-    }, [config.PluginSettings.Plugins, defaultValue, id, setInitialSetting]);
+    }, [config.PluginSettings.Plugins, id, setInitialSetting]);
 
     const optionOnChangeHandler = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
         const value = e.target.name === 'surveyEnabled';
--- a/webapp/src/components/systemConsole/enableSurvey/index.tsx
+++ b/webapp/src/components/systemConsole/enableSurvey/index.tsx
@@ -1,0 +1,58 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useCallback, useEffect, useState} from 'react';
+
+import type {CustomSettingChildComponentProp} from 'components/systemConsole/index';
+
+import './style.scss';
+
+function EnableSurvey({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) {
+    const defaultValue = false;
+
+    const [enabled, setEnabled] = useState<boolean>(defaultValue);
+
+    useEffect(() => {
+        const enabledSurveyConfig = config.PluginSettings.Plugins['com.mattermost.user-survey']?.systemconsolesetting.SurveyEnabled;
+
+        const initialValue = enabledSurveyConfig || defaultValue;
+        setEnabled(initialValue);
+        setInitialSetting(id, initialValue);
+    }, [config.PluginSettings.Plugins, defaultValue, id, setInitialSetting]);
+
+    const optionOnChangeHandler = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+        if (e.target.name === 'surveyEnabled') {
+            setEnabled(true);
+        } else if (e.target.name === 'surveyDisabled') {
+            setEnabled(false);
+        }
+    }, []);
+
+    return (
+        <div className='EnableSurvey horizontal'>
+            <div className='option horizontal'>
+                <input
+                    type='radio'
+                    name='surveyEnabled'
+                    id='enableSurvey_enabled'
+                    checked={enabled}
+                    onChange={optionOnChangeHandler}
+                />
+                <label htmlFor='enableSurvey_enabled'>{'Enabled'}</label>
+            </div>
+
+            <div className='option horizontal'>
+                <input
+                    type='radio'
+                    name='surveyDisabled'
+                    id='enableSurvey_disabled'
+                    checked={!enabled}
+                    onChange={optionOnChangeHandler}
+                />
+                <label htmlFor='enableSurvey_disabled'>{'Disabled'}</label>
+            </div>
+        </div>
+    );
+}
+
+export default EnableSurvey;
--- a/webapp/src/components/systemConsole/enableSurvey/index.tsx
+++ b/webapp/src/components/systemConsole/enableSurvey/index.tsx
@@ -21,11 +21,7 @@ function EnableSurvey({id, setSaveNeeded, onChange, config, setInitialSetting}: 
     }, [config.PluginSettings.Plugins, defaultValue, id, setInitialSetting]);
 
     const optionOnChangeHandler = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
-        let value = false;
-        if (e.target.name === 'surveyEnabled') {
-            value = true;
-        }
-
+        const value = e.target.name === 'surveyEnabled';
         setEnabled(value);
         onChange(id, value);
         setSaveNeeded();
--- a/webapp/src/components/systemConsole/enableSurvey/index.tsx
+++ b/webapp/src/components/systemConsole/enableSurvey/index.tsx
@@ -13,7 +13,7 @@ function EnableSurvey({id, setSaveNeeded, onChange, config, setInitialSetting}: 
     const [enabled, setEnabled] = useState<boolean>(defaultValue);
 
     useEffect(() => {
-        const enabledSurveyConfig = config.PluginSettings.Plugins['com.mattermost.user-survey']?.systemconsolesetting.SurveyEnabled;
+        const enabledSurveyConfig = config.PluginSettings.Plugins['com.mattermost.user-survey']?.systemconsolesetting.EnableSurvey;
 
         const initialValue = enabledSurveyConfig || defaultValue;
         setEnabled(initialValue);
@@ -21,12 +21,15 @@ function EnableSurvey({id, setSaveNeeded, onChange, config, setInitialSetting}: 
     }, [config.PluginSettings.Plugins, defaultValue, id, setInitialSetting]);
 
     const optionOnChangeHandler = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+        let value = false;
         if (e.target.name === 'surveyEnabled') {
-            setEnabled(true);
-        } else if (e.target.name === 'surveyDisabled') {
-            setEnabled(false);
+            value = true;
         }
-    }, []);
+
+        setEnabled(value);
+        onChange(id, value);
+        setSaveNeeded();
+    }, [id, onChange, setSaveNeeded]);
 
     return (
         <div className='EnableSurvey horizontal'>
--- a/webapp/src/components/systemConsole/expiry/expiry.tsx
+++ b/webapp/src/components/systemConsole/expiry/expiry.tsx
@@ -10,11 +10,10 @@ import type {ExpiryConfig} from 'types/mattermost-webapp';
 
 import './style.scss';
 
-const Expiry = ({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) => {
-    // default values
-    const defaultExpiry = '30';
+const DEFAULT_SURVEY_EXPIRY = '30';
 
-    const [expiryDays, setExpiryDays] = useState<string>(defaultExpiry);
+const Expiry = ({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) => {
+    const [expiryDays, setExpiryDays] = useState<string>(DEFAULT_SURVEY_EXPIRY);
     const [error, setError] = useState<string>('');
 
     // Set initial value from saved config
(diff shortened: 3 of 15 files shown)


webapp/src/components/systemConsole/expiry/expiry.tsx
```diff
@@ -4,25 +4,34 @@
...
 
-        const expiryConfig = config.PluginSettings?.Plugins?.['com.mattermost.user-survey']?.surveyexpiry;
+        const initialSetting: ExpiryConfig = {
+            days: Number.parseInt(defaultExpiry, 10),
+        };
 
         if (expiryConfig?.days) {
             setExpiryDays(expiryConfig.days.toString());
+            initialSetting.days = expiryConfig.days;
         }
```
larkox, line 31: In the previous component, we set the local value always, but here we only set the local state if the config is truthy (so we wouldn't update it if the config is 0).

Is this intended?

webapp/src/components/systemConsole/expiry/expiry.tsx
```diff
@@ -4,25 +4,34 @@
...
-    }, [config.PluginSettings?.Plugins]);
+
+        setInitialSetting(id, initialSetting);
+    }, [config.PluginSettings?.Plugins, id, setInitialSetting]);
```
larkox, line 34: Same question as in previous component about id and setinitialsetting.

webapp/src/components/systemConsole/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+            ...existingConfig,
+            [settingId]: settings,
+        }));
+    }, []);
+
+    const modifiedProps = useMemo((): CustomSettingChildComponentProp => ({
+        ...props,
+        onChange: onChangeWrapper,
+        setInitialSetting: setDefaults,
+    }), [onChangeWrapper, props, setDefaults]);
```
larkox, line 55: You are expanding the object, so in terms of memoization this is not needed. If you did it in terms of removing stress for the garbage collector, I could think it deeper whether it makes sense or not.

webapp/src/components/systemConsole/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    //
+    // This is needed so that on change in one sub-setting, we know all other sub-settings as well and thus
+    // can save the entire setting object on server side.
+    // Otherwise the displayed default values will never get saved unless a user changes them.
+    const setDefaults = useCallback((settingId: string, settings: CustomConfigTypes) => {
+        setConfig((existingConfig) => ({
+            ...existingConfig,
+            [settingId]: settings,
+        }));
+    }, []);
```
larkox, line 49: In order to simplify all this logic we could have this component (or another file) store all the defaults, and just use that. So oversimplifying everything...

constants.tsx
```
const DEFAULT_CONFIG = {
    key1: 'value1'
}
```
index.tsx:
```
import DEFAULT_CONFIG from './constants';

function mergeDefaultConfig(stateConfig) {
    const mergedConfig = {};
    for (key in DEFAULT_CONFIG) {
        mergedConfig[key] = stateConfig[key] ?? DEFAULT_CONFIG[key];
    }
    return mergedConfig;
}
const SystemConsoleSetting = (props) => {
    const [config, setConfig] = useState(() => mergeDefaultConfig(props.config....));
    const onChangeWrapper = useCallback((...) => {
        setConfig((existingConfig) => ({
            ...existingConfig,
            [settingId]: settings,
        }));
     }, []);

    ...
}
```
settingX.tsx:
```
import DEFAULT_CONFIG from './constants`
const SettingX = (props) => {
    const [localValue, setLocalValue] = useState(props.config..... ?? DEFAULT_CONFIG.settingX);
    ...
}
```

webapp/src/components/systemConsole/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+                <div className='settingLabel'>
+                    {'Enable survey:'}
+                </div>
+                <div className='customSettingComponent'>
+                    <EnableSurvey
+                        {...modifiedProps}
+                        id='EnableSurvey'
+                    />
+                </div>
+            </div>
```
larkox, line 69: Nit: This is pattern is repeated here for every component, so in theory it could be extracted into a component, or into a loop, or something to not repeat code. For example...

```
const settings = [{title: 'Enable survey:', component: (<EnableSurvey {...modifiedProps} id='EnableSurvey' />)}, {title: 'Send next survey at:', component: (<SurveyDateTime .../>)}, ...];
const body = settings.map((s) => (
       <div className='horizontal'>
                <div className='settingLabel'>
                    {s.title}
                </div>

                <div className='customSettingComponent'>
                    {s.component}
                </div>
            </div>
);
```
Or...
```
const SettingWrapper = ({title, children}) => (
     <div className='horizontal'>
                <div className='settingLabel'>
                    {title}
                </div>

                <div className='customSettingComponent'>
                    {children}
                </div>
            </div>
);


function SystemConsoleSetting(props: CustomComponentProps) {
    ...
    return (
        <div className='SystemConsoleSetting vertical'>
            <SettingWrapper title = 'Enable survey:'>
                    <SurveyDateTime
                        {...modifiedProps}
                        id='SurveyDateTime'
                    />
            </SettingWrapper>
            <SettingWrapper title='Send next survey at:'>
                ...
            </SettingWrapper>
            ...
        </div>
    );
}
```

But up to you whether it is worth it making any change like that.

webapp/src/components/systemConsole/questions/questions.tsx
```diff
@@ -0,0 +1,147 @@
...
+                mandatory: true,
+            },
+            {
+                id: utils.uuid(),
+                type: 'text',
+                system: false,
+                mandatory: false,
+            },
+        ];
+    };
```
larkox, line 65: I don't see any reason to have this function inside the component (and being regenerated on every render).

I have mixed feelings on whether it needs to be a function at all. The only reason I see is that we may want to have the ids different every time we call this, but not sure if that is actually needed.

webapp/src/components/systemConsole/questions/questions.tsx
```diff
@@ -0,0 +1,147 @@
...
+                ...newQuestions[i],
+                text: e.target.value,
+            };
+            setQuestions(newQuestions);
+            saveSettings(newQuestions);
+        }, [questions, saveSettings]),
+        500,
+    );
+
+    const renderQuestions = useMemo(() => {
```
larkox, line 95: Nit: `renderX` sounds like an action, therefore a function. Maybe `renderedQuestions`? 0/5

Naming is hard.

webapp/src/components/systemConsole/questions/questions.tsx
```diff
@@ -0,0 +1,147 @@
...
+        return questions.map((question) => {
+            return (
+                <div
+                    key={question.id}
+                    className='vertical question'
+                >
+                    <span className='questionTitle'>
+                        {question.title && question.title}
```
larkox, line 103: I see what you are doing here, but it is really confusing. Why not moving this logic outside of the render like:

```
title = question.title ? question.title : (`${questionTypeDisplayName.get(question.type)} ${question.mandatory ? '' : '(Optional)'}`);
return (
    <div ...>
        <span className='questionTitle'>
            {title}
        </span>
        ...
   </div>
);
```

Maybe making it a bit different so it is better readable.

webapp/src/components/systemConsole/teamFilter/teamFilter.tsx
```diff
@@ -53,26 +54,25 @@ function TeamFilter({id, setSaveNeeded, onChange, config}: CustomComponentProps)
...
-        }
-    ), []);
+    const customComponents: CustomComponentsDefinition = useMemo(() => ({
+        Option: CustomOption,
+        MultiValueContainer: CustomMultiValueContainer,
+    }), []);
+
+    const optionsToConfig = (teams: DropdownOption[]): TeamFilterConfig => ({
+        filteredTeamIDs: teams.map((option) => option.value),
+    });
```
larkox, line 71: This doesn't depend on any prop or state, so it can be moved outside of the component.

larkox (approved):
Looks great, but you know how I am. A bunch of nitpicks and so on in case you want to address them, but I am ok with the code shipping as is.

webapp/src/components/systemConsole/enableSurvey/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+// See LICENSE.txt for license information.
+
+import React, {useCallback, useEffect, useState} from 'react';
+
+import type {CustomSettingChildComponentProp} from 'components/systemConsole/index';
+
+import './style.scss';
+
+function EnableSurvey({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) {
+    const defaultValue = false;
```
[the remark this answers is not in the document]
devinbinnie: Yeah I'd agree this does seem unnecessary unless we're planning on changing it.

webapp/src/components/systemConsole/enableSurvey/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    const [enabled, setEnabled] = useState<boolean>(defaultValue);
+
+    useEffect(() => {
+        const enabledSurveyConfig = config.PluginSettings.Plugins['com.mattermost.user-survey']?.systemconsolesetting.EnableSurvey;
+
+        const initialValue = enabledSurveyConfig || defaultValue;
+        setEnabled(initialValue);
+        setInitialSetting(id, initialValue);
+    }, [config.PluginSettings.Plugins, defaultValue, id, setInitialSetting]);
```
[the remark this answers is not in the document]
devinbinnie: I'm guessing the linter was complaining that these values which are in the `useEffect` callback are not in the dependencies - perhaps we need to adjust the linter to be less opinionated about this?

webapp/src/components/systemConsole/questions/questions.tsx
```diff
@@ -0,0 +1,147 @@
...
+    helpText?: string;
+};
+
+function SurveyQuestions({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) {
+    const [questions, setQuestions] = useState<Question[]>([]);
+
+    const generateDefaultQuestions = (): Question[] => {
+        return [
+            {
+                id: utils.uuid(),
```
devinbinnie, line 36: Why `uuid` out of curiosity? Do we just want a unique id here for these? If so, could we just hardcode them as default somehow? Or is there some validation where we want these to be UUIDs?

webapp/src/components/systemConsole/questions/questions.tsx
devinbinnie: Yeah not sure why we're `&&`ing this. If it's blank/null it just won't get rendered right?

webapp/src/components/systemConsole/questions/questions.tsx
```diff
@@ -0,0 +1,147 @@
...
+                        maxLength={1000}
+                        className={`form-control questionInput ${question.system && 'disabled'}`}
+                        defaultValue={question.text}
+                        onChange={(e) => questionOnChangeHandler(e, question.id)}
+                        disabled={question.system}
+                        placeholder='Question'
+                    />
+
+                    {
+                        question.helpText &&
```
devinbinnie, line 120: Nit: can we put this on the above line? Just to save some space. I know this is for readibility, but this is a bit overkill IMO. 0/5 though.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -2,64 +2,76 @@
...
 
-import type {CustomComponentProps} from 'types/mattermost-webapp';
+import type {DateTimeConfig} from 'types/mattermost-webapp';
 
 import './style.scss';
 
-const defaultSurveyTime = '09:00';
+function SurveyDateTime({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) {
+    // Default settings
+    const defaultSurveyTime = '09:00';
```
devinbinnie, line 17: Same comment as Daniel has for constants, they should be defined outside the function.

webapp/src/utils/utils.tsx
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// Source: https://www.robots.ox.ac.uk/~adutta/blog/standalone_uuid_generator_in_javascript.html
+function uuid() {
```
devinbinnie, line 5: There's a UUID library we could use instead of this - might be worth using that instead of re-defining it ourselves

webapp/src/components/systemConsole/questions/questions.tsx
```diff
@@ -0,0 +1,147 @@
...
+        setInitialSetting(id, initialSetting);
+    }, [config.PluginSettings.Plugins, id, setInitialSetting]);
+
+    const saveSettings = useCallback((setting: SurveyQuestionsConfig) => {
+        setSaveNeeded();
+        onChange(id, setting);
+    }, [id, onChange, setSaveNeeded]);
+
+    const questionOnChangeHandler = useDebouncedCallback(
+        useCallback((e: React.ChangeEvent<HTMLInputElement>, questionID: string) => {
```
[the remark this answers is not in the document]
harshilsharma63: You're right, it already memomizes the function so no need of `useCallback`. Done.

webapp/package.json
```diff
@@ -53,6 +53,7 @@
     "classnames": "2.5.1",
     "core-js": "3.36.0",
     "date-fns": "3.3.1",
+    "debounce-promise": "3.1.2",
```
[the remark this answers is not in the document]
harshilsharma63: Removed. It was the previous not-so-nice debounce library I added, so removed it now.

webapp/src/components/systemConsole/enableSurvey/index.tsx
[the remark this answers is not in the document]
harshilsharma63: I'm using it in two places right now so want to have the default value in a constant to serve as a single place of truth instead of the value being spread around.

Agree on moving it outside the component as uppercase snake case. Done!

webapp/src/components/systemConsole/enableSurvey/index.tsx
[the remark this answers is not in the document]
harshilsharma63: Devin is right, its to make the linter happy.

webapp/src/components/systemConsole/expiry/expiry.tsx
```diff
@@ -4,25 +4,34 @@
...
+import type {CustomSettingChildComponentProp} from 'components/systemConsole/index';
+
+import type {ExpiryConfig} from 'types/mattermost-webapp';
 
 import './style.scss';
 
-const defaultExpiry = '30';
+const Expiry = ({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) => {
+    // default values
+    const defaultExpiry = '30';
```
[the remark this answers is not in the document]
harshilsharma63: Done. Changed the case and moved outside the component.

webapp/src/components/systemConsole/expiry/expiry.tsx
harshilsharma63: Not sure where you're referring, but I'm initializing the state with default value already so need of doing it here.

I think you're referring to the `SurveyDateTime` component, there I'm setting state even if false value in config as the default value os not computable in a single line, or rather it cleaner in multiple lines.

webapp/src/components/systemConsole/expiry/expiry.tsx
harshilsharma63: Making the linter happy.

webapp/src/components/systemConsole/index.tsx
harshilsharma63: It won't work when the default value itself is dynamic. For example the survey date's default value is today + 30 days. Later when we add the option to select survey cadence, the default value of date and time will change based on that setting.

I think its more flexible if each component managed its own default config for now as the requirement and scope can change. We're still in the POC stage kinda.

webapp/src/components/systemConsole/index.tsx
harshilsharma63: Done.

webapp/src/components/systemConsole/questions/questions.tsx
harshilsharma63: generating different IDs was part of the reason. While I don't know what problem it causes if IDs weren't re-generated everytime we needed this, I had a feeling it might cause issues. Its a function because if its an array inside the component, it will be generated on every render. It can either be a useMemo, or a function like this.

webapp/src/components/systemConsole/questions/questions.tsx
harshilsharma63: Makes sense. Done.

webapp/src/components/systemConsole/questions/questions.tsx
harshilsharma63: Okay, so the part about `question.title` and `question.helpText` are removed now as they were used by the survey message text, which is not a question. I misinterpreted it as a question and added it in the array of questions. So took it out of questions array and thus removed title and helpText fields from question type. So this part of code is gone now.

webapp/src/components/systemConsole/teamFilter/teamFilter.tsx
harshilsharma63: Done.

webapp/src/components/systemConsole/questions/questions.tsx
harshilsharma63: Just need something nice and random. This will also be needed when we add the functionality to add custom questions.

webapp/src/components/systemConsole/questions/questions.tsx
harshilsharma63: This code is gone now.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: Done.

webapp/src/utils/utils.tsx
devinbinnie: For sure, but I know there is a library that also does this (and probably the same way). It also would mean we wouldn't be on the hook if the implementation needed to change for security reasons (which is also a double edged sword). Just something to think about, 0/5 on actually changing it.

devinbinnie (approved):

PR #8, opened 2024-04-09
User survey post component

harshilsharma63:

### DIFF
--- a/webapp/src/components/common/button/button.tsx
+++ b/webapp/src/components/common/button/button.tsx
@@ -6,22 +6,27 @@ import React, {useMemo} from 'react';
 
 import './style.scss';
 
-export type Props = {
+export interface Props extends React.ButtonHTMLAttributes<HTMLButtonElement> {
     text?: string;
-    type?: 'primary' | 'secondary' | 'tertiary';
+    buttonType?: 'primary' | 'secondary' | 'tertiary';
     danger?: boolean;
     iconClass?: string;
     iconPlacement?: 'left' | 'right';
-    disabled?: boolean;
-    onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
-};
-function Button({text, type = 'tertiary', danger = false, iconClass, iconPlacement = 'left', disabled = false, onClick}: Props) {
+}
+function Button({
+    text,
+    buttonType = 'tertiary',
+    danger = false,
+    iconClass,
+    iconPlacement = 'left',
+    ...rest
+}: Props) {
     const buttonClassName = classNames({
         btn: true,
         iconButton: !text,
-        'btn-primary': type === 'primary',
-        'btn-secondary': type === 'secondary',
-        'btn-tertiary': type === 'tertiary',
+        'btn-primary': buttonType === 'primary',
+        'btn-secondary': buttonType === 'secondary',
+        'btn-tertiary': buttonType === 'tertiary',
         'btn-danger': danger,
     });
 
@@ -37,8 +42,7 @@ function Button({text, type = 'tertiary', danger = false, iconClass, iconPlaceme
     return (
         <button
             className={buttonClassName}
-            disabled={disabled}
-            onClick={onClick}
+            {...rest}
         >
             {iconClass && iconPlacement === 'left' && iconComponent}
 
--- a/webapp/src/components/common/button/button.tsx
+++ b/webapp/src/components/common/button/button.tsx
@@ -12,8 +12,10 @@ export type Props = {
     danger?: boolean;
     iconClass?: string;
     iconPlacement?: 'left' | 'right';
+    disabled?: boolean;
+    onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void
 };
-function Button({text, type = 'tertiary', danger = false, iconClass, iconPlacement = 'left'}: Props) {
+function Button({text, type = 'tertiary', danger = false, iconClass, iconPlacement = 'left', disabled = false, onClick}: Props) {
     const buttonClassName = classNames({
         btn: true,
         iconButton: !text,
@@ -33,7 +35,11 @@ function Button({text, type = 'tertiary', danger = false, iconClass, iconPlaceme
     }, [iconClass]);
 
     return (
-        <button className={buttonClassName}>
+        <button
+            className={buttonClassName}
+            disabled={disabled}
+            onClick={onClick}
+        >
             {iconClass && iconPlacement === 'left' && iconComponent}
 
             {text}
--- a/webapp/src/components/common/button/button.tsx
+++ b/webapp/src/components/common/button/button.tsx
@@ -13,7 +13,7 @@ export type Props = {
     iconClass?: string;
     iconPlacement?: 'left' | 'right';
     disabled?: boolean;
-    onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void
+    onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
 };
 function Button({text, type = 'tertiary', danger = false, iconClass, iconPlacement = 'left', disabled = false, onClick}: Props) {
     const buttonClassName = classNames({
--- a/webapp/src/components/hooks/survey.ts
+++ b/webapp/src/components/hooks/survey.ts
@@ -55,8 +55,8 @@ export function useUserSurvey(surveyID: string) {
 
         setSurvey(survey);
 
-        // the first linear scale question is the system default rating question
-        const linearScaleQuestion = survey.questions.find((question) => question.type === 'linear_scale');
+        // the first system, linear scale question is the system default rating question
+        const linearScaleQuestion = survey.questions.find((question) => question.system && question.type === 'linear_scale');
 
         if (linearScaleQuestion) {
             linearScaleQuestionID.current = linearScaleQuestion.id;
@@ -68,16 +68,17 @@ export function useUserSurvey(surveyID: string) {
     }, [surveyID]);
 
     const setResponses = useCallback((responses: SurveyResponse) => {
-        if (!survey) {
-            return;
-        }
+        setSurvey((oldSurvey) => {
+            if (!oldSurvey) {
+                return oldSurvey;
+            }
 
-        const newSurvey: UserSurvey = {
-            ...survey,
-            response: responses,
-        };
-        setSurvey(newSurvey);
-    }, [survey]);
+            return {
+                ...oldSurvey,
+                response: responses,
+            };
+        });
+    }, []);
 
     return {
         survey,
--- a/webapp/src/components/hooks/survey.ts
+++ b/webapp/src/components/hooks/survey.ts
@@ -1,0 +1,89 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {useCallback, useEffect, useRef, useState} from 'react';
+
+import type {SurveyResponse, UserSurvey} from 'types/plugin';
+
+export function useUserSurvey(surveyID: string) {
+    const [survey, setSurvey] = useState<UserSurvey>();
+    const linearScaleQuestionID = useRef<string>();
+
+    const responsesExist = survey?.response !== undefined;
+    const surveyExpired = survey?.status === 'ended';
+
+    const fetchUserSurvey = async (surveyID: string) => {
+        // make API call here to get user survey...
+        const survey: UserSurvey = {
+            surveyId: surveyID,
+            startDate: '01/04/2024',
+            endDate: '30/04/2024',
+            questions: [
+                {
+                    id: '49fc9a85-b4d8-424e-b13f-40344b168123',
+                    mandatory: true,
+                    system: true,
+                    text: 'How likely are you to recommend Mattermost?',
+                    type: 'linear_scale',
+                },
+                {
+                    id: 'aa026055-c97c-48d7-a025-c44590078963',
+                    mandatory: true,
+                    system: true,
+                    text: 'How can we make your experience better?',
+                    type: 'text',
+                },
+                {
+                    id: '0eee4429-5083-41ef-bda7-2ee9c5ece929',
+                    mandatory: false,
+                    system: false,
+                    text: 'option question text',
+                    type: 'text',
+                },
+            ],
+            status: 'in_progress',
+
+            // response: {
+            //     responses: {
+            //         '49fc9a85-b4d8-424e-b13f-40344b168123': '8',
+            //         'aa026055-c97c-48d7-a025-c44590078963': 'Response 1',
+            //         '0eee4429-5083-41ef-bda7-2ee9c5ece929': 'Response 2',
+            //     },
+            //     dateCreated: format(new Date(), 'dd/MM/yyyy'),
+            // },
+        };
+
+        setSurvey(survey);
+
+        // the first linear scale question is the system default rating question
+        const linearScaleQuestion = survey.questions.find((question) => question.type === 'linear_scale');
+
+        if (linearScaleQuestion) {
+            linearScaleQuestionID.current = linearScaleQuestion.id;
+        }
+    };
+
+    useEffect(() => {
+        fetchUserSurvey(surveyID);
+    }, [surveyID]);
+
+    const setResponses = useCallback((responses: SurveyResponse) => {
+        if (!survey) {
+            return;
+        }
+
+        const newSurvey: UserSurvey = {
+            ...survey,
+            response: responses,
+        };
+        setSurvey(newSurvey);
+    }, [survey]);
+
+    return {
+        survey,
+        linearScaleQuestionID,
+        responsesExist,
+        surveyExpired,
+        setResponses,
+    };
+}
--- a/webapp/src/components/surveyPost/linearScaleQuestion/linearScaleQuestion.tsx
+++ b/webapp/src/components/surveyPost/linearScaleQuestion/linearScaleQuestion.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import classNames from 'classnames';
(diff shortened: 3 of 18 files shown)


webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -76,3 +78,46 @@ export type SurveyResult = {
...
+    status: SurveyStatus;
+}
+
+export type Question = {
+    id: string;
+    text?: string;
+    type: QuestionType;
+    system: boolean;
+    mandatory: boolean;
+};
```
larkox, line 119: This file should define types coming from the `mattermost-webapp` package, isn't it? These don't seem to be part of the mattermost webapp package. Shouldn't they live in a different file?

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+            const newDraftResponse: SurveyResponse = draftResponse.current || {
+                responses: {},
+                dateCreated: format(new Date(), 'dd/MM/yyyy'),
+            };
+            newDraftResponse.responses[questionID] = response;
+            draftResponse.current = newDraftResponse;
+
+            // if this is the system rating question, submit response ASAP
+            if (questionID === linearScaleQuestionID.current) {
+                submitRating();
```
larkox, line 100: If the user keeps changing the value over and over we keep sending this over and over. I guess this is expected and handled somewhere else.

larkox (approved):
LGTM, a few comments, but everything minor.

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+        };
+
+        // send payload to submit survey response API here...
+    };
+
+    const renderedMessage = useMemo(() => {
+        // @ts-expect-error window is definitely MattermostWindow in plugins
+        const mmWindow = window as MattermostWindow;
+        const htmlString = mmWindow.PostUtils.formatText(post.message, {markdown: true});
+        return mmWindow.PostUtils.messageHtmlToComponent(htmlString, isRHS);
```
hmhealey, line 154: `isRHS` is actually no longer used, so you're safe to remove it

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+            let questionComponent: React.ReactNode;
+
+            switch (question.type) {
+            case 'linear_scale':
+                questionComponent = (
+                    <LinearScaleQuestion
+                        question={question}
+                        responseChangeHandler={questionResponseChangeHandler}
+                        disabled={disabled}
+                        value={survey.response?.responses[question.id] as number}
```
hmhealey, line 172: Are all of these props expected to be the same for any type of question? If each question component has the same props, you could rewrite this switch statement using an object to map between the `type` and the corresponding component. I don't know how TS will handle this, but we've used this pattern elsewhere
```js
const questionComponents = {
    linear_scale: LinearScaleQuestion,
    text: TextQuestion,
};

const Question = questionComponents[question.type];

return (
    <Question
        question={question}
        responseChangeHandler={questionResponseChangeHandler}
        disabled={disabled}
        value={survey.response?.responses[question.id} // This is the line I don't know about
    />
);
```

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+function SurveyPost({post, isRHS}: CustomPostTypeComponentProps) {
+    const [survey, setSurvey] = useState<UserSurvey>();
+    const [disabled, setDisabled] = useState<boolean>(false);
+    const [expired, setExpired] = useState<boolean>(false);
+    const draftResponse = useRef<SurveyResponse>();
+    const [errorMessage, setErrorMessage] = useState<string>();
+
+    const linearScaleQuestionID = useRef<string>();
+
+    const fetchUserSurvey = async (surveyID: string) => {
```
hmhealey, line 29: 0/5, but perhaps it would be a good idea to factor out all the stuff around managing the survey state into a hook so that it could be tested independently of the component rendering?

webapp/src/components/surveyPost/textQuestion/style.scss
```diff
@@ -0,0 +1,8 @@
+.TextQuestion {
+    .questionTitle {
+        color: #3D3C40;
```
hmhealey, line 3: Is this going to be visible on different themes?

webapp/src/components/surveyPost/textQuestion/textQuestion.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function TextQuestion({question, responseChangeHandler, disabled, value}: Props) {
+    const changeHandler = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+        responseChangeHandler(question.id, e.target.value);
+    }, [question.id, responseChangeHandler]);
+
+    return (
+        <div className='TextQuestion vertical'>
+            <div className='questionTitle'>{question.text}</div>
```
hmhealey, line 24: Could this be written using a label to improve accessibility?
```html
<label>
    {question.text}
    <input/>
</label>
```
That should make it so that, when a user navigates to the input with a screen reader, it should read out the question to them

hmhealey (changes requested):

webapp/src/components/common/button/button.tsx
```diff
@@ -12,8 +12,10 @@ export type Props = {
     danger?: boolean;
     iconClass?: string;
     iconPlacement?: 'left' | 'right';
+    disabled?: boolean;
+    onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
 };
-function Button({text, type = 'tertiary', danger = false, iconClass, iconPlacement = 'left'}: Props) {
+function Button({text, type = 'tertiary', danger = false, iconClass, iconPlacement = 'left', disabled = false, onClick}: Props) {
```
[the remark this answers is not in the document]
harshilsharma63: Done.

webapp/src/components/surveyPost/linearScaleQuestion/linearScaleQuestion.tsx
```diff
@@ -0,0 +1,60 @@
...
+export type Props = {
+    question: Question;
+    responseChangeHandler: (questionID: string, response: number) => void;
+    disabled?: boolean;
+    value?: number;
+    scaleStart?: number;
+    scaleEnd?: number;
+}
+
+function LinearScaleQuestion({question, responseChangeHandler, disabled, value, scaleStart = 1, scaleEnd = 10}: Props) {
```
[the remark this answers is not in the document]
harshilsharma63: Done.

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+        const surveyExpired = survey.status === 'ended';
+
+        setDisabled(responsesExist || surveyExpired);
+        setExpired(surveyExpired);
+    };
+
+    // fetch data on initial mount
+    useEffect(() => {
+        fetchUserSurvey(post.props.surveyID);
+    }, [post.props]);
```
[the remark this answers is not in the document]
harshilsharma63: yes yes, its fine.

webapp/src/types/mattermost-webapp/index.d.ts
harshilsharma63: Ah yes, you're right, its under the mm webapp dir. I've created a new `plugin.d.ts` file now and moved plugin specific types in there.

webapp/src/components/common/button/button.tsx
```diff
@@ -12,8 +12,10 @@ export type Props = {
     danger?: boolean;
     iconClass?: string;
     iconPlacement?: 'left' | 'right';
+    disabled?: boolean;
```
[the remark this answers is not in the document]
harshilsharma63: Great idea! I didn't know about that! Although I had to use `React.ButtonHTMLAttributes<HTMLButtonElement>` instead to give the correct button attributes.

webapp/src/components/surveyPost/surveyPost.tsx
harshilsharma63: Yes, this is expected and we'll handle this on the server side.

harshilsharma63: Maybe the linear scale's change handler should be debounced for slight longer...

webapp/src/components/surveyPost/linearScaleQuestion/style.scss
```diff
@@ -0,0 +1,75 @@
...
+            border-radius: 16px;
+            background: rgba(var(--button-bg-rgb), 0.08);
+            gap: 8px;
+
+            &.disabled {
+                background: var(--center-channel-color-8, rgba(63, 67, 80, 0.08));
+                pointer-events: none;
+
+                .indent.selected {
+                    background: var(--center-channel-color-72, rgba(63, 67, 80, 0.72));
```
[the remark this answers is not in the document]
harshilsharma63: Why does it matter and what difference would it make?

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+        // the first linear scale question is the system default rating question
+        const linearScaleQuestion = survey.questions.find((question) => question.type === 'linear_scale');
+        linearScaleQuestionID.current = linearScaleQuestion?.id;
+
+        const responsesExist = survey.response !== undefined;
+        const surveyExpired = survey.status === 'ended';
+
+        setDisabled(responsesExist || surveyExpired);
+        setExpired(surveyExpired);
```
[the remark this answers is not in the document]
harshilsharma63: yes, it works! there was no need to store those in state individually. Done!

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+    };
+
+    // fetch data on initial mount
+    useEffect(() => {
+        fetchUserSurvey(post.props.surveyID);
```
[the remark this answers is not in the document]
harshilsharma63: I thought about this but realistically there shouldn't be a large number of survey posts. I'm thinking about storing the response in the post props itself on submission so it doesn't need to make those API calls, but yeah, something that we could do. But I don't think its a concern.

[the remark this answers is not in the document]
harshilsharma63: The part I'm a bit worried about is that the API calls are made every time you switch to the DM.

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+    }, [post.props]);
+
+    const questionResponseChangeHandler = useDebouncedCallback(
+        (questionID: string, response: unknown) => {
+            const newDraftResponse: SurveyResponse = draftResponse.current || {
+                responses: {},
+                dateCreated: format(new Date(), 'dd/MM/yyyy'),
```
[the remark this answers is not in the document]
harshilsharma63: Its fine and expected as the date and time the admin selects for the survey is UTC, so there will be some users on -1 day and some on the day and maybe some on +1 day. This is fine as we're setting the timezone expectation with sysadmins. As for the user, they see the date they submitted the survey on, whatever day it was in their time zone, so from their POV the date is correct.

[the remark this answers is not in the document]
harshilsharma63: Maybe if the earth was flat we could have gone with the whole world having the same timezone 🤔 🤔 🤔

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,243 @@
...
+            //         '0eee4429-5083-41ef-bda7-2ee9c5ece929': 'Response 2',
+            //     },
+            //     dateCreated: format(new Date(), 'dd/MM/yyyy'),
+            // },
+        };
+
+        setSurvey(survey);
+
+        // the first linear scale question is the system default rating question
+        const linearScaleQuestion = survey.questions.find((question) => question.type === 'linear_scale');
```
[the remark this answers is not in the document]
harshilsharma63: Just in case we add a different first question. Checking for the question type feels a bit assuring 😛

webapp/src/components/surveyPost/surveyPost.tsx
harshilsharma63: Done.

webapp/src/components/surveyPost/linearScaleQuestion/style.scss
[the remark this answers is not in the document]
harshilsharma63: Discussed this yesterday with @hmhealey and the reason is that we eventually wanna get rid of the `-rgba` color variables and entirely depend on the primary colors to simplify the color pallett. So, its a good idea to start not using those color variables as they'll be gone sometime in the future.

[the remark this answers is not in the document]
harshilsharma63: @hmhealey I'll fix this as part of https://mattermost.atlassian.net/browse/MM-57748

webapp/src/components/surveyPost/surveyPost.tsx
harshilsharma63: Nice idea! It worked! Done

webapp/src/components/surveyPost/surveyPost.tsx
harshilsharma63: As in creating a custom hook to fetch the survey?

webapp/src/components/surveyPost/textQuestion/style.scss
harshilsharma63: It don't. I have a separate ticket to fix the colors here and in system console. I'll create a separate PR to fix the colors.

webapp/src/components/surveyPost/textQuestion/textQuestion.tsx
harshilsharma63: Done.

webapp/src/components/surveyPost/surveyPost.tsx
harshilsharma63: Done. Created `useUserSurvey` hook. Please check it out. This is my first official custom hook 🎉 🎉 🎉

webapp/src/components/common/button/button.tsx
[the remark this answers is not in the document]
hmhealey: I hate that the built-in types have weird, slightly different versions of props like that. I guess that's just something you have to deal since all these types are old

webapp/src/components/surveyPost/surveyPost.tsx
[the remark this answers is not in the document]
hmhealey: That's unfortunately something that we don't have a general solution for yet. For stuff in Redux, we assume that objects are always kept in sync via the websocket or we assume that they're always stale and need to be refetched like this.

I agree that it's probably fine in this case though since it only affects that one DM channel

webapp/src/components/surveyPost/surveyPost.tsx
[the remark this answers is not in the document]
hmhealey: Perhaps we could add a special flag to that one survey question to mark that it's the general NPS question? That's only needed if we plan to let users reorder the questions though so that the NPS one isn't first

webapp/src/components/hooks/survey.ts
```diff
@@ -0,0 +1,89 @@
...
+    const setResponses = useCallback((responses: SurveyResponse) => {
+        if (!survey) {
+            return;
+        }
+
+        const newSurvey: UserSurvey = {
+            ...survey,
+            response: responses,
+        };
+        setSurvey(newSurvey);
```
hmhealey, line 79: Instead of passing `survey` into this, I found out that the callback version of the old `setState` works with `useState` hooks as well:
```js
setSurvey((prevSurvey) => {
    if (!prevSurvey) {
        return prevSurvey;
    }

    return {
        ...prevSurvey,
        response: responses,
    };
);
```
Not that it matters in this case, but it can also help if you happen to have multiple calls to `setSurvey` at once since each one will always be called with the latest value of `survey`

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const disabled = responsesExist || surveyExpired;
+
+    const submitSurveyHandler = useCallback(async () => {
+        if (!draftResponse.current ||
+            Object.keys(draftResponse.current?.responses).length === 0
+        ) {
+            return;
+        }
+
+        const response = await submitSurveyResponse();
```
hmhealey, line 46: I was thinking that the logic for submitting the survey could also be part of that hook as well. My goal is to get as much of the logic around managing the survey into the hook and to keep the component to just rendering the results.

If that ends up being too much though, feel free to ignore this one. I know you've got other stuff to work on with this, and I don't want to bog you down by getting obsessed with hooks suddenly 😛

hmhealey (changes requested):

webapp/src/components/surveyPost/surveyPost.tsx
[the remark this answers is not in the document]
harshilsharma63: Added check for system type question along with question type.

webapp/src/components/hooks/survey.ts
harshilsharma63: Makes sense. Done!

webapp/src/components/surveyPost/surveyPost.tsx
hmhealey: Yep, totally fine

hmhealey (approved):

PR #10, opened 2024-04-12
Server tests

harshilsharma63:
#### Summary
1. Added store test framework to run tests on MySQL and Postgres.
2. Added migration tests.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-57833


devinbinnie (approved):

larkox (approved):

go.mod
```diff
@@ -5,54 +5,98 @@ go 1.21
 require (
 	github.com/mattermost/mattermost/server/public v0.1.0
 	github.com/mattermost/morph v1.1.0
+	github.com/pborman/uuid v1.2.1
```
esarafianou, line 8: @harshilsharma63 Any reason we used this package instead of Google's? https://pkg.go.dev/github.com/google/uuid.

Google's `NewString` seems to do what `NewRandom` of this package does. 

I'd prefer that we use Google's that's well maintained.

PR #11, opened 2024-04-23
Added scheduled job to manage survey status

harshilsharma63:
#### Summary
1. Added a scheduled job to manage survey status - to end in progress survey if needed, and to start new survey if needed.
2. Added SQL migration functions to add and remove columns in an idempotent fashion.
3. Added HTTP API endpoint framework and created a dummy API to be implemented next.
4. Created Store interface and added mockery to generate interface mocks

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-57986


server/app/survey.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (a *UserSurveyApp) GetInProgressSurvey() (*model.Survey, error) {
+	surveys, err := a.store.GetSurveysByStatus(model.SurveyStatusInProgress)
+	if err != nil {
+		return nil, errors.Wrap(err, "GetInProgressSurvey: failed to get in progress surveys from database")
+	}
+
+	if len(surveys) > 1 {
+		return nil, fmt.Errorf("more than one in-progress survey found in the database. There should only be one in-progress survey, in_progress_survey_count: %d", len(surveys))
+	}
```
larkox, line 31: Don't we want to give more information here? Maybe the surveys ids so a survey can be stopped? 0/5

server/job.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (p *Plugin) startScheduledJobs() error {
+	if err := p.startSurveyStartJob(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Plugin) startSurveyStartJob() error {
```
larkox, line 28: It is intended that this name has `start` twice?

server/job.go
```diff
@@ -0,0 +1,44 @@
...
+	startSurveyJonInterval = 15 * time.Second
+)
+
+func (p *Plugin) startScheduledJobs() error {
+	if err := p.startSurveyStartJob(); err != nil {
+		return err
+	}
+
+	return nil
+}
```
larkox, line 26: This is just a wrapper over `startSurveyStartJob`. Do we want to keep it separate to add more stuff in the future?

Also, in its current shape, even if we want to keep different names, we can do just
```suggestion
func (p *Plugin) startScheduledJobs() error {
	return p.startSurveyStartJob()
}
```

server/store/migrations/000003_add_survey_fields.up.sql
```diff
@@ -0,0 +1,2 @@
+{{- /* addColumnIfNeeded tableName columnName datatype constraint */ -}}
```
larkox, line 1: Is this needed? I understand it is documentation, but 0/5 if this is the place for it.

webapp/package.json
```diff
@@ -3,7 +3,7 @@
   "scripts": {
     "build": "webpack --mode=production",
     "build:watch": "webpack --mode=production --watch",
-    "debug": "webpack --mode=none",
+    "debug": "webpack --mode=production",
```
larkox, line 6: On debug we want production mode? If so, wouldn't we want also production mode in `debug:watch`? 0/5

webapp/src/components/systemConsole/questions/questions.tsx
```diff
@@ -22,7 +22,7 @@ const DEFAULT_SURVEY_MESSAGE_TEXT = 'Please take a few moments to help us improv
 
 function SurveyQuestions({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) {
     const [questions, setQuestions] = useState<Question[]>([]);
-    const [surveyMessageText, setSurveyMessageText] = useState<string>(DEFAULT_SURVEY_MESSAGE_TEXT);
+    const [surveyMessageText, setSurveyMessageText] = useState<string>();
```
larkox, line 25: I don't get the changes in this file. Why these changes?

server/store/templateHelperFuncs.go
```diff
@@ -0,0 +1,108 @@
...
+			"schema":          s.schemaName,
+			"table_name":      tableName,
+			"norm_table_name": normalizedTableName,
+			"column_name":     columnName,
+			"data_type":       dataType,
+			"constraint":      constraint,
+		}
+
+		x := replaceVars(`
+			SET @stmt = (SELECT IF(
```
larkox, line 29: Cannot we use query builders for this (and other stuff on this file)?

larkox (approved):
LGTM, just a few questions

server/app/survey.go
harshilsharma63: Done!

server/job.go
harshilsharma63: ah, my bad. Renamed to `JobManageSurveyStatus` now.

webapp/src/components/systemConsole/questions/questions.tsx
harshilsharma63: I'm using `surveyMessageText` as a input `defaultValue`. The way react deals with input placeholder is that it doesn't re-render the inputs on change in `defaultValue`. So, if I initialise the state with `DEFAULT_SURVEY_MESSAGE_TEXT `, then later in a userEffect I change it to a value from the saved config, the input isn't updated as it should based on documentation. So, I removed the default value from state variable and let the `useEffect` set either the default or the saved value from config.

Without this, the saved value from config never shows up.

server/store/templateHelperFuncs.go
harshilsharma63: Its unnecessarily complex to use query builder for this. Also, Gorm doesn't even support these stuff, we end up using raw queries anyways. We went through this idea in focalboard as well, where this code is copied from 😛

webapp/package.json
harshilsharma63: Ah no, this should be `development`.

server/store/migrations/000003_add_survey_fields.up.sql
harshilsharma63: Nah, not needed, its just a habit from Focalboard days 😛 I've removed it now.

server/job.go
harshilsharma63: I kept it this way because I thought we'd need two jobs - one to start a survey and one to end, but a single job is sufficient and better for it, so `startScheduledJobs` is unnecessary now. I've removed it now and can add it later when there is an actual need for it.

devinbinnie (approved):

harshilsharma63:
Merging as @esarafianou is out for next week as well, so will fix any comments later on.

server/store/templateHelperFuncs.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/mattermost/mattermost-plugin-user-survey/server/model"
+)
+
+func (s *SQLStore) genAddColumnIfNeeded(tableName, columnName, dataType, constraint string) (string, error) {
+	tableName = addPrefixIfNeeded(tableName, s.tablePrefix)
+	normalizedTableName := s.normalizeTablename(tableName)
+
+	switch s.dbType {
+	case model.DBTypeMySQL:
```
esarafianou, line 18: @harshilsharma63 Why are we adding MySQL support if we're deprecating it? Same question in the `genDropColumnIfNeeded` function

server/app/survey.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if len(surveys) > 1 {
+		var surveyIDs string
+		for _, survey := range surveys {
+			surveyIDs += " " + survey.ID
+		}
+
+		return nil, fmt.Errorf("more than one in-progress survey found in the database. There should only be one in-progress survey, in_progress_survey_count: %d, surveyIDs: %s", len(surveys), surveyIDs)
+	}
```
esarafianou, line 36: @harshilsharma63 Not a security comment, but considering the checks we perfrom in `JobManageSurveyStatus` before we check for a new survey to start, how can we end up with more than one survey in progress?

esarafianou: Plus, if we don't allow two concurrent surveys, shouldn't we check for it before we save a new survey?

server/app/job.go
```diff
@@ -0,0 +1,100 @@
...
+			CreateAt:        now,
+			UpdateAt:        now,
+			StartTime:       startTime.UnixMilli(),
+			Duration:        config.SurveyExpiry.Days,
+			SurveyQuestions: config.SurveyQuestions,
+			Status:          model.SurveyStatusInProgress,
+		}
+
+		a.api.LogDebug("JobManageSurveyStatus: saving new survey")
+		err = a.SaveSurvey(survey)
```
esarafianou, line 89: @harshilsharma63 I'm a bit confused. It seems like we save a survey when it's about to start. How would then an admin create a survey to be started at a future date/time?

server/store/templateHelperFuncs.go
harshilsharma63: Unfortunately some customers are still on MySQL and we want them to use this plugin.

server/app/job.go
harshilsharma63: When you tweak the plugin system console settings, all changes are saved in `config.json` or wherever Mattermost is storing its config. This is the default mechanism of Mattermost to save a plugin's config. So, when you select a date, time or configure survey questions, those are all being saved by Mattermost like any other plugin setting. We bring stuff into the plugin-created tables only when a survey starts as before that, we have nothing to do with it.

server/app/survey.go
harshilsharma63: There is no code path that would allow more than one in-progress survey. This check is just here because I'm receiving an array of surveys and thought why not check for it anyways. Its especially important in the initial days of this plugin while its still being tested.

We do ensure that an in-progress survey, if one exists, is stopped (if needed) before starting a new one (if needed).

PR #12, opened 2024-04-25
Send survey

harshilsharma63:
#### Summary
1. Implemented `/connected` API to send the survey if needed.
2. Added logic to call the API when plugin's webapp component is loaded.
3. 

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-57886


webapp/src/components/hooks/survey.ts
```diff
@@ -3,56 +3,23 @@
...
-            //     },
-            //     dateCreated: format(new Date(), 'dd/MM/yyyy'),
-            // },
-        };
+    useEffect(() => {
+        if (!post.props.survey_questions) {
+            return;
+        }
 
+        const survey = JSON.parse(post.props.survey_questions) as Survey;
```
larkox, line 22: This is a security issue waiting to happen.

Post props are not validated around the app, so we need to add some validation to `survey` so there is not a crash when trying to access its values.

I would say the simplest is just to consider this instead of Survey, consider it `Partial<Survey>`. But we could do some more in depth verification.

Also... JSON.parse can throw an exception if the string is not a valid JSON, so it would be better to wrap it on a try catch.

larkox (changes requested):
A couple of thoughts, and a thing to look into.

server/api/utils.go
```diff
@@ -0,0 +1,36 @@
...
+package api
+
+import (
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+func (api *Handlers) RequireAuthentication(w http.ResponseWriter, r *http.Request) error {
+	userID := r.Header.Get(headerMattermostUserID)
```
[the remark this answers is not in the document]
harshilsharma63: So all the plugin we have, Focalboard, MS teams, Github etc, simply do whats being done here, so I think its fine.

server/app/survey.go
```diff
@@ -50,3 +63,183 @@ func (a *UserSurveyApp) StopSurvey(surveyID string) error {
...
+		Type:      surveyPostType,
+	}
+
+	questionsJSON, err := json.Marshal(survey.SurveyQuestions)
+	if err != nil {
+		a.api.LogError("SendSurvey: failed to marshal survey questions for inserting into post", "error", err.Error())
+		return errors.Wrap(err, "SendSurvey: failed to marshal survey questions for inserting into post")
+	}
+
+	post.AddProp("survey_questions", string(questionsJSON))
```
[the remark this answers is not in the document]
harshilsharma63: In its current form, yes, mobile users won't see the survey but rather only the `Message` part of the post. The idea is to get started on the mobile side development once the plugin is ready. But yes, until then mobile users can't submit the survey.

webapp/src/components/hooks/survey.ts
harshilsharma63: Done, added generated type checking code in `validators.ts`.

harshilsharma63: @larkox I actually added the check in the other PR as there were a lot of  changes with the hook in that PR and it avoided re-doing some work. Its in #13 validators.ts -

harshilsharma63: apologies for the confistion.

larkox (approved):
Approving based on the comments answers.

server/app/survey.go
```diff
@@ -50,3 +64,197 @@ func (a *UserSurveyApp) StopSurvey(surveyID string) error {
...
+	if appErr != nil {
+		a.api.LogError("setSurveySentToUser: Failed to set user survey sent status in KV store", "userID", userID, "surveyID", surveyID, "error", appErr.Error())
+		return errors.Wrap(appErr, "setSurveySentToUser: Failed to set user survey sent status in KV store")
+	}
+
+	return nil
+}
+
+func (a *UserSurveyApp) SendSurvey(userID string, survey *model.Survey) error {
+	if err := a.ensureSurveyBot(); err != nil {
```
hmhealey, line 129: Could we do this when the plugin is initialized instead of waiting until the survey is sent?

server/app/survey_test.go
```diff
@@ -0,0 +1,204 @@
...
+	t.Run("base case - empty survey should fail", func(t *testing.T) {
+		th := SetupAppTest(t)
+
+		err := th.App.SaveSurvey(&model.Survey{})
+		require.Error(t, err, "should return error as empty survey is invalid")
+	})
+
+	t.Run("should successfully save a valid survey", func(t *testing.T) {
+		th := SetupAppTest(t)
+		th.MockedStore.On("SaveSurvey", mock.Anything).Return(nil)
```
hmhealey, line 28: I'm kind of surprised that we mock the store here given that that's part of the plugin. I know you'd need to do that for causing error cases, but I wouldn't think you'd need it for success cases.

Is that just for these tests since it involves error cases, or will you always need to mock the store?

webapp/src/client/httpClient.ts
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Client4} from 'mattermost-redux/client';
+import {ClientError} from 'mattermost-redux/client/client4';
+
+class HttpClient {
```
hmhealey, line 7: 0/5 but would it make sense for this to extend `Client4` so that you could use `doFetch` instead of reimplementing that? You wouldn't have the URL set automatically by the web app, but you'd get the rest of that as-is

webapp/src/selectors/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const getServerRoute = (state: any) => {
+    const config = getConfig(state);
+
+    let basePath = '';
+    if (config && config.SiteURL) {
+        basePath = new URL(config.SiteURL).pathname;
+
+        if (basePath && basePath[basePath.length - 1] === '/') {
+            basePath = basePath.substr(0, basePath.length - 1);
```
hmhealey, line 14: `.substr` is deprecated. I forget the difference between it and `.susbstring`, but you should probably use that instead

hmhealey (changes requested):
Looks pretty good. Most of the things that actually require changes are the comments related to locking and concurrency

server/api/utils.go
```diff
@@ -0,0 +1,36 @@
...
+		http.Error(w, "Unauthenticated", http.StatusUnauthorized)
+		return errors.New("Unauthenticated user")
+	}
+
+	return nil
+}
+
+func (api *Handlers) DisallowGuestUsers(w http.ResponseWriter, r *http.Request) error {
+	userID := r.Header.Get(headerMattermostUserID)
+	user, appErr := api.pluginAPI.GetUser(userID)
```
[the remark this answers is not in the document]
harshilsharma63: Yeah, I get your point, but I don't think there is any other way without changing MM server logic or plugin APIs.

server/app/lock.go
```diff
@@ -0,0 +1,40 @@
...
+		msg := fmt.Sprintf("TryLock: failed to save value in KV store via KVCompareAndSet, key: %s, value: %s, error: %s", key, timeJSON, appErr.Error())
+		a.api.LogError(msg)
+		return false, errors.New(msg)
+	}
+
+	return locked, nil
+}
+
+func (a *UserSurveyApp) Unlock(key string) error {
+	appErr := a.api.KVDelete(key)
```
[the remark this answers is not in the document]
harshilsharma63: Nice catch! Fixed now.

server/app/lock.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func (a *UserSurveyApp) TryLock(key string, now time.Time) (bool, error) {
+	timeJSON, err := json.Marshal(now)
+	if err != nil {
+		a.api.LogError("TryLock: failed to marshal time value", "value", now.String())
+		return false, errors.Wrap(err, "TryLock: failed to marshal time value")
+	}
+
+	locked, appErr := a.api.KVCompareAndSet(key, nil, timeJSON)
```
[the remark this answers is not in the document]
harshilsharma63: Done. Referred the `clearStaleLocks` logic from NPS plugins.

server/app/survey.go
```diff
@@ -50,3 +64,197 @@ func (a *UserSurveyApp) StopSurvey(surveyID string) error {
...
+	if err != nil {
+		return false, errors.Wrap(err, "ShouldSendSurvey: failed to acquire lock for checking if survey needs to be sent to user or not")
+	}
+
+	defer func() {
+		_ = a.Unlock(key)
+	}()
+
+	if !locked {
+		return false, nil
```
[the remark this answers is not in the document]
harshilsharma63: Yup, done.

server/app/survey.go
```diff
@@ -50,3 +64,197 @@ func (a *UserSurveyApp) StopSurvey(surveyID string) error {
 
 	return nil
 }
+
+func (a *UserSurveyApp) ShouldSendSurvey(userID string, survey *model.Survey) (bool, error) {
+	key := utils.KeyUserSendSurveyLock(userID)
+	locked, err := a.TryLock(key, time.Now().UTC())
```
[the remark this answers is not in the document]
harshilsharma63: Very insightful! I've refactored this to move the lock acquisition and release process in API layer, however I'm not having it span the duration for checking and sending, rather only the sending function, which should be enough as it performs multiple database calls and creates posts, so should be enough.

The primary reason why I don't want to acquire lock before checking is that if we do so, every time any user loads up MM, we'll create a KV store entry and delete it, for every single MM user on the server, every time they load MM. So, right now, with lock acquisition logic only before the general SendSurvey function, two threads will both see that they both need to send the survey, but only only one will be able to acquire the lock before sending, so the end result is the same minus two DB queries.

Do you feel this might not be sufficient and we still need to expand the lock duration to "check and send"?

server/app/survey.go
```diff
@@ -50,3 +64,197 @@ func (a *UserSurveyApp) StopSurvey(surveyID string) error {
...
+	alreadySent, err := a.getSurveySentToUser(userID, survey.ID)
+	if err != nil {
+		return false, errors.Wrap(err, "ShouldSendSurvey: failed to check if survey is already sent to the user")
+	}
+
+	if alreadySent {
+		return false, nil
+	}
+
+	inExcludedTeam, err := a.userInExcludedTeams(userID, survey)
```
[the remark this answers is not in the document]
harshilsharma63: Done. With the above mentioned refactored, this is automatically handled.

server/app/survey.go
harshilsharma63: Done.

server/app/survey_test.go
harshilsharma63: Generally I just mock store layer in app layer's unit tests. Store later is supposed to have its own unit tests against read database, but at the moment they don't exist (the framework for adding them does, though). So the idea is at some point I add store layer tests as well, and combined, the two unit tests would be complete. Does that resolve your concern?

webapp/src/client/httpClient.ts
harshilsharma63: I don't think that class is exported, but yeah, will check it out later.

webapp/src/selectors/index.tsx
harshilsharma63: Done.

server/app/survey.go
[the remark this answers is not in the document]
hmhealey: No, that's still not quite enough because we could still end up with the following series of events that'd result in the user getting two surveys:
1. Goroutine A checks and sees that the survey hasn't been sent
2. Goroutine A acquires the lock
3. Goroutine B checks and sees that the survey hasn't been sent
4. Goroutine A sends the survey and releases the lock
5. Goroutine B acquires the lock, sends the survey, and releases the lock

You don't need to hold the lock to check if the there's a survey open, but you do need to hold it before you check if the survey has already been sent to avoid that. I suppose that you could check if the survey has been sent twice,(once without the lock and then once with it) if you think that trying to acquire the lock will have a performance hit, but I don't expect that to have much of an impact. That also only could have a performance impact if there's a survey open as well, so it's even less likely that there's any effect if people aren't constantly running surveys

server/app/survey_test.go
hmhealey: Personally, I tend to prefer not using mocks where possible because it covers more of the code at once and it does it in more of a realistic way. I ran into that with the MS Teams plugin testing where, with mocks, it felt like I was basically testing things like "calling `saveSomething` should call `store.SaveSomething`" which confirms that the code doesn't contain typos, but not that it did what I want it to. With a real store, I could change that to "calling `saveSomething` sets a value in the DB and returns another value" because I'm not having to pretend how `store.SaveSomething` behaves. Hopefully that example makes some sense and isn't too hand wavey 😅

Also, I tend to find mocking is a bit of a pain to do which is another reason I avoid it when possible

webapp/src/client/httpClient.ts
hmhealey: It should be available from `@mattermost/client` instead of getting it from `mattermost-redux`

server/app/survey_test.go
harshilsharma63: Consider these as unit tests. I can add the integration tests later as per time. Created https://mattermost.atlassian.net/browse/MM-58103 to track this.

webapp/src/client/httpClient.ts
harshilsharma63: Will do later if needed in interest of time.

harshilsharma63:
@hmhealey I'll address the integration test thing later, everything else should be good.

server/app/survey_test.go
hmhealey: That works, and sorry, I didn't mean to sound like we needed to change all these now. I've just recently liked talking about tests and how I think they can be done better

hmhealey (changes requested):
I hate how GitHub collapses stuff sometimes since I think you missed my latest response about the TOCTOU bugs. See https://github.com/mattermost/mattermost-plugin-user-survey/pull/12#discussion_r1588475238 since I think there's still potential for a race here

server/api/utils.go
```diff
@@ -0,0 +1,36 @@
...
+
+package api
+
+import (
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+func (api *Handlers) RequireAuthentication(w http.ResponseWriter, r *http.Request) error {
```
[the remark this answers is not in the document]
harshilsharma63: yeah, that would be ideal

server/app/survey.go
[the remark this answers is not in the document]
harshilsharma63: Makes sense. fixed now! Acquiring lock after check if there is an in progress survey, and releasing it after sending the survey post.

server/app/survey_test.go
harshilsharma63: No worries, I got your point.

harshilsharma63:
> I hate how GitHub collapses stuff sometimes since I think you missed my latest response about the TOCTOU bugs. See [#12 (comment)](https://github.com/mattermost/mattermost-plugin-user-survey/pull/12#discussion_r1588475238) since I think there's still potential for a race here

Fixed now. Yeah, its not a great UX when there are new comments on resolved threads :(

hmhealey (approved):

server/api/connection.go
```diff
@@ -5,9 +5,77 @@ package api
...
+	if err := api.RequireAuthentication(w, r); err != nil {
+		return
+	}
+
+	if err := api.DisallowGuestUsers(w, r); err != nil {
+		return
+	}
+
+	// check if there is an in progress survey, if so,
+	// check if the user has already been sent to the user, if not,
```
esarafianou, line 23: nit: `check if the **survey** has already been sent to the user, if not,

server/app/survey.go
```diff
@@ -50,3 +64,211 @@ func (a *UserSurveyApp) StopSurvey(surveyID string) error {
...
+	}
+
+	userMemberOfFilteredTeam := false
+
+	for _, team := range teams {
+		if _, ok := filteredTeamMap[team.Id]; ok {
+			userMemberOfFilteredTeam = true
+			break
+		}
+	}
```
esarafianou, line 242: @harshilsharma63 this seems to imply that if a user is a member of teams: Team A, Team B, Team C and only Team C is on the excluded teams, they will not receive the survey despite being in teams A and B that are not excluded. Is that the desired state?

harshilsharma63: @esarafianou your'e absolutely right, they won't receive the survey. This is indeed the desired behaviour.

This was discussed here as well for more context - https://hub.mattermost.com/private-core/pl/ncrx1rchfjbbjb8hrdatc3qony

esarafianou: Sounds good, thanks for the context!

PR #15, opened 2024-05-07
Added slash command to reset all user survey data for debugging purpose

harshilsharma63:
#### Summary
Added a slash command to delete all user survey data for debugging purpose. It deletes all responses, all metadata about whether a survey was sent to a user or not, etc.

The command is registered only if the environment variable `MM_SERVICESETTINGS_ENABLEDEVELOPER` is enabled at build time. So, by default, the plugin builds won't include this slash command, as expected.

The purpose of this is to use the command while testing so one can keep testing surveys after resetting the data.

The slash command is `/resetdata`. It reports the error if there was an error, else prints a success message.

Only system admins are permitted to run this command.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-58144

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ endif
 	mkdir -p server/dist;
 ifneq ($(MM_SERVICESETTINGS_ENABLEDEVELOPER),)
 	@echo Building plugin only for $(DEFAULT_GOOS)-$(DEFAULT_GOARCH) because MM_SERVICESETTINGS_ENABLEDEVELOPER is enabled
-	cd server && env CGO_ENABLED=0 $(GO) build $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-$(DEFAULT_GOOS)-$(DEFAULT_GOARCH);
+	cd server && env CGO_ENABLED=0 $(GO) build -ldflags '-X main.DebugBuild=true' $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-$(DEFAULT_GOOS)-$(DEFAULT_GOARCH);
 else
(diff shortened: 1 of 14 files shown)


M-ZubairAhmed (approved):
Dont we have tests for these?

Makefile
```diff
@@ -102,7 +102,7 @@ endif
 	mkdir -p server/dist;
 ifneq ($(MM_SERVICESETTINGS_ENABLEDEVELOPER),)
 	@echo Building plugin only for $(DEFAULT_GOOS)-$(DEFAULT_GOARCH) because MM_SERVICESETTINGS_ENABLEDEVELOPER is enabled
-	cd server && env CGO_ENABLED=0 $(GO) build $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-$(DEFAULT_GOOS)-$(DEFAULT_GOARCH);
+	cd server && env CGO_ENABLED=0 $(GO) build -ldflags '-X main.DebugBuild=true' $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-$(DEFAULT_GOOS)-$(DEFAULT_GOARCH);
```
devinbinnie, line 105: Is this just temporary or should this always be on?

devinbinnie (approved):

Makefile
harshilsharma63: This will always be on if `MM_SERVICESETTINGS_ENABLEDEVELOPER` env variable is set at build time. In production build CI env, this variable will not be set and so the slash command won't work. I'll use this to tweak some more behaviours in production/dev builds.

harshilsharma63:
> Dont we have tests for these?

At this point no, but the idea is to get the reporting finished and get started with QA, UX review, text review etc and I'll be adding a lot of tests while that happens.

PR #20, opened 2024-05-14
Fixd UUID code to generate 26 letter IDs as standard

harshilsharma63:
#### Summary
#19 caused a bug as it generated UUIDs that were longer than 26 letters. Since DB columns were designed for 26 letter IDs, it broke the store layer. This PR includes some changes to use Google UUID library to generate 26 letter UUIDs.


hmhealey (approved):

agnivade (approved):
Approving for now. But some comments to consider for later:
- Improve the model.NewId implementation and use it everywhere.
- Add a test case to ensure the 26 char length.

harshilsharma63:
@agnivade I've added a test now.

PR #22, opened 2024-05-14
DB constraints and indexes

harshilsharma63:
#### Summary
Added btree indexes on a few columns which are commonly queried.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-58176


server/store/migrations/000012_survey_table_constraints_and_indexes.down.sql
```diff
@@ -0,0 +1 @@
+SELECT 1;
```
larkox, line 1: Why a select and not dropping the indexes?

larkox (approved):
LGTM, just one question

server/store/migrations/000012_survey_table_constraints_and_indexes.down.sql
harshilsharma63: Ideally we should delete the index, but we never did it in Focalboard, not sure why though. Since I was just referencing Boards for this, I used the same pattern.

harshilsharma63: I'll try to remember if there was any reason for this or it "just happened".

devinbinnie: Yeah I would definitely be in favour of actually dropping them, we were fine doing it in the server at least: https://github.com/mattermost/mattermost/blob/master/server/channels/db/migrations/postgres/000118_create_index_poststats.down.sql#L1

devinbinnie (approved):

devinbinnie (changes requested):
Just want to clear up the index dropping first.

server/store/migrations/000012_survey_table_constraints_and_indexes.down.sql
harshilsharma63: @devinbinnie its easy to do in Postgres, like in the file you linked. If you see the corresponding migration file for MySQL, we're doing nothing there - https://github.com/mattermost/mattermost/blob/master/server/channels/db/migrations/mysql/000118_create_index_poststats.down.sql

Its fine for now and should not be a blocking issue. Generating the query for this and testing is time consuming. Nt worth it for a down migration.

harshilsharma63: Nevermind, found MySQL queries in newer Mattermost server migrations. Updating now.

harshilsharma63: Done. Added drop index SQL now @devinbinnie @larkox

PR #24, opened 2024-05-15
UI refresh

harshilsharma63:
#### Summary
Updated UI and test as per latest Figma designs

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-58238


abhijit-singh (changes requested):
Great work @harshilsharma63! I had a few requests after an initial review:
1. For the `Enable plugin` field, we should show all the config UI as soon as the user selects `true`. The user should not have to click on Save for the fields to show up.  If the plugin is initially disabled, in order to get a survey started, the user has to enable the plugin and hit save. Then the fields actually show and the user has to setup the survey settings and save again to apply. This will be much smoother if we just allow them to configure the survey when they select true in the top field and then they can hit save once. Another reason is that there is some hesitation to enable plugin without knowing what will happen when they hit save. If we show them all the fields when they select true, they will know what all they can configure and setup and hence help reduce that hesitation.
2. We should not allow people to select a past date/time for the survey. To keep things simple, in the date field, we can only allow them to select a future date and not the current or past date.  Also, in the edge case where a date is selected but the `Enable survey` field is set to false, and then after that date has passed, the admin comes in and changes that field to `true` with a past date still selected, we should add a validation and throw an error asking them to select a future date. 
image

3. The calendar for the date picker field is getting cut. It should be able to extend beyond that container. 
image

4. For the Survey expiry field, can we convert it into a numeric field so that random text is not allowed? Along with that, we’ll also need a validation to make sure it is not set to a negative value. We can trigger this validation as soon as the user focuses away from that field with an invalid value entered.
image

5. For the help text below the time and date field, there are a few different states of it based on if values are selected or not. Can we update that to make sure the user understands what they’re setting up?
image



There are other minor cosmetic issues:
1. The fonts in the section title need to be updated. There also seems to be some extra space between the title and the subtitle that can be reduced. Lastly, the arrow is slightly misaligned vertically and needs to move up by a few pixels so that space above and below it is the same. Here’s the [design for reference](https://www.figma.com/design/Q7seEbiQboa5GGClUcDHqD/NPS-Plugin-v2?node-id=2636-88205&t=R5cYVgAlqmuT90ZY-4).
2. The options for the `Enable survey` field should not be in bold.
3. It’ll be nice to vertically centre align the field labels with the input block. 
image

4. In the Active and Past surveys section, can we add an arrow next to start time to inform the user how the list is sorted. It doesn’t have to be interactive though. 
5. In the Active and Past surveys section, can we vertically centre align the content in the rows?
image

M-ZubairAhmed:
I will review once UX is completed, please add me then

harshilsharma63:
@abhijit-singh 

1. Not possible in plugins, its just how plugins work.
2. Done. I've disabled selecting past or current date. A validation error is still pending and would likely addressed later as it isn't turning out to be a simple solution or I'm missing something.
3. Fixed. 
4. Done.
5. Done except the part about showing when was the last survey sent at. That info wasn't present in the date picker component. LMK if its a blocker and I'll figure it out.
6. minor cosmetic issues - done.

abhijit-singh:
Thanks @harshilsharma63. The functional updates are working as expected. 

A few UI issues still seem to be there: 
1. (Carried over from before) In the Active and Past surveys section, can we vertically centre align the content in the rows?
image
2. In the section headers, there's an extra 10px of margin-bottom applied to the subtitle text which is making it vertically misaligned. We can get rid of that margin for the section header subtitles. 
image
3. Also, we need a 4px space between the section header title and subtitle as per the design. A margin bottom of 4px on the title does it.
image
4. The collapse/expand arrow indicator is in the opposite direction of what it should be. Point down while collapsed and pointing up when expanded. Similar to how it is [shown here.](https://mui.com/material-ui/react-accordion/)

Coming to the user facing side, there are few cosmetic issues here as well:
1. In the linear scale question, the numbers don't seem to be vertically center aligned and need to be moved down a bit (they're slightly shifted to the top). Changing their line height to 26px fixed it for me. 
Current:
image
Ideal:
image
2. For the selected state of the ratings 1 and 10 (first and last), can we make it a complete circle like it is for the others? I believe setting the border radius to 16px irrespective of which position it is on will fix it.
image
3. For the text "Not Likely" and "Very Likely", can we change the font-weight to 600.
4. Lets change the text input boxes to text-area inputs as discussed. 
5. Can we reduce the space between the field label and the actual input to 8px? Currently it seems to be 12 px. 
image

6. Lastly, can we increase the space between the submit button and the last input box to 20px? It is sticking together a little too much right now. 
image

harshilsharma63:
@abhijit-singh I've finished all UI changes we've discussed over the last few days. Please re-review. I've also updated the plugin in the test server.

abhijit-singh (approved):
Looks awesome @harshilsharma63! Thank you!

Just one small thing - In the `Scheduled` view of the section header, the 4px space between the title and subtitle is missing. You can also notice the UI moving up and down a bit when selecting the date because of this. Can we add that to make things consistent?

harshilsharma63:
> Looks awesome @harshilsharma63! Thank you!
> 
> Just one small thing - In the `Scheduled` view of the section header, the 4px space between the title and subtitle is missing. You can also notice the UI moving up and down a bit when selecting the date because of this. Can we add that to make things consistent?

Fixed!

devinbinnie (approved):

PR #25, opened 2024-05-16
Checked for duplicate survey before starting a new one

harshilsharma63:
#### Summary
Don't start new survey if latest ended survey and the config are the same.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-58241

### DIFF
--- a/server/app/job.go
+++ b/server/app/job.go
@@ -67,6 +67,15 @@ func (a *UserSurveyApp) startNewSurveyIfNeeded() error {
 	}
 
 	if shouldStartSurvey {
+		// compare with latest ended survey if it's not the same as the survey we're trying to start.
+		// When an admin ends a survey manually within its duration, not checking this can result
+		// in the job starting a duplicate survey immoderately. So, compare the config with last ended survey
+		// and only start a new survey if they both are different.
+		endedSurvey, err := a.store.GetLatestEndedSurvey()
+		if err != nil {
+			return errors.Wrap(err, "startNewSurveyIfNeeded: failed to get latest ended survey from database")
+		}
+
 		a.api.LogDebug("JobManageSurveyStatus: determined that the new survey should start")
 		now := mmModal.GetMillis()
 		startTime, err := config.ParsedTime()
@@ -74,7 +83,7 @@ func (a *UserSurveyApp) startNewSurveyIfNeeded() error {
 			return errors.Wrap(err, "JobManageSurveyStatus: failed to read survey parsed time")
 		}
 
-		survey := &model.Survey{
+		surveyFromConfig := &model.Survey{
 			ID:              utils.NewID(),
 			ExcludedTeamIDs: config.TeamFilter.FilteredTeamIDs,
 			CreateAt:        now,
@@ -87,12 +96,17 @@ func (a *UserSurveyApp) startNewSurveyIfNeeded() error {
 
 		for _, question := range config.SurveyQuestions.Questions {
 			if question.Text != "" {
-				survey.SurveyQuestions.Questions = append(survey.SurveyQuestions.Questions, question)
+				surveyFromConfig.SurveyQuestions.Questions = append(surveyFromConfig.SurveyQuestions.Questions, question)
 			}
 		}
 
+		if surveyFromConfig.IsEqual(endedSurvey) {
+			a.api.LogDebug("JobManageSurveyStatus: not starting new survey as it is the same as latest ended survey")
+			return nil
+		}
+
 		a.api.LogDebug("JobManageSurveyStatus: saving new survey")
-		err = a.SaveSurvey(survey)
+		err = a.SaveSurvey(surveyFromConfig)
 		if err != nil {
 			return errors.Wrap(err, "JobManageSurveyStatus: failed to save survey in database")
 		}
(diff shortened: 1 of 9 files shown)


larkox (approved):
Codewise LGTM, but I have a UX question.

If I end a survey because whatever reason, and I want to start the same survey again, how would I do that? I have to change something so it starts again?

yasserfaraazkhan (commented):
image

@harshilsharma63 just a curious thought.
would it be easier to unset the Date field after a survey is launched? rather than comparing survey settings data.

**Scenario 1**: If there is an on going survey, and user needs to schedule a new one.

User can select a future date, and wait for current Survey to expire on its own. This automatically launches a survey. (✅ working currently)

**Scenario 2**: If user wants to end the going survey.

Since the Date field is empty. System can't launch a new survey. User can end the current on going survey prematurely. This prevents unwanted surveys from launching automatically. For new survey the Flow will be as in Scenario 1.

**Scenario 3**: (w.r.t future nps work) If there's a cadence option like quarterly, half yearly

- For surveys scheduled at specific intervals (e.g., quarterly, half-yearly), I guess we will automatically calculate the launch date based on the selected cadence.
- If a user ends a survey before its scheduled end date, then we adjusts the launch date for the next survey accordingly.
- This ensures that there are no duplicate surveys launched automatically, even if a survey is terminated earlier than planned.

harshilsharma63:
@yasserfaraazkhan thanks a ton for the feedback. We can do anything and everything given sufficient time 😉

Scheduling of surveys is not a one-solution-fits-all and 10 users will suggest 10 different ways to do it, so we need to act on our customer's feedback first before spending time in changes which we may not even want.

> would it be easier to unset the Date field after a survey is launched? rather than comparing survey settings data.

We can update the logic to not start new survey if "Enable Survey" is set to false. I don't want to make a field empty-able unless absolutely needed and some logic might need changes around it.

> Since the Date field is empty. System can't launch a new survey. User can end the current on going survey prematurely. This prevents unwanted surveys from launching automatically. For new survey the Flow will be as in Scenario 1.

This scenario if fixed in this PR. Previously if you ended a survey prematurely, anew one would launch automatically which is exactly what this PR fixes. With this PR, a new survey will launch only if its different from the most recently ended survey in terms of questions, excluded teams, duration or start date-time.

> For surveys scheduled at specific intervals (e.g., quarterly, half-yearly), I guess we will automatically calculate the launch date based on the selected cadence.

Yes,

> If a user ends a survey before its scheduled end date, then we adjusts the launch date for the next survey accordingly.

This is not decided yet. There are too many scenarios and too many options for this so this is best left to decide later after receiving some user feedback.

harshilsharma63:
> If I end a survey because whatever reason, and I want to start the same survey again, how would I do that? I have to change something so it starts again?

Yes, you'd have to change something.

Up to @abhijit-singh how we want to handle it but I'd refrain from making such changes before receiving real-world user feedback as there are too many options and no right or wrong option.

abhijit-singh:
I'm okay with having to change something to trigger the survey for now because I think it'll be a super rare edge case. A couple of questions: 
1. How difficult would it be to show this as an error to the user instead of silently failing? We can show a modal explaining that they already sent the exact same survey out.  
2. @harshilsharma63 We discussed not letting admins start a survey with the past date/time selected. It is blocked on the UI already, but if it is pre-selected and the admin comes to make a change, we discussed we'll show an error on the date field and ask them to select a future date. Where are we on that?

harshilsharma63:
> I'm okay with having to change something to trigger the survey for now because I think it'll be a super rare edge case. A couple of questions:
> 
> 1. How difficult would it be to show this as an error to the user instead of silently failing? We can show a modal explaining that they already sent the exact same survey out.
> 2. @harshilsharma63 We discussed not letting admins start a survey with the past date/time selected. It is blocked on the UI already, but if it is pre-selected and the admin comes to make a change, we discussed we'll show an error on the date field and ask them to select a future date. Where are we on that?

1. Can you define exactly what condition to show the modal and where and on performing what action?

2. The form validation didn't work and I spent lot of time on it. I'll be working on other core bugs first before picking up the form validation. The other part of UI changes are other other than fixing fonts, margins, padding etc so I'd say we can merge the existing UI fixes and deal with remaining later.

harshilsharma63:
@abhijit-singh I see that a log of UX issues are associated with the inability to clear out the date field. If thats the case, let me work on that part as well. But can you think all cases and sceneries to see if we're missing anything else or there are more weird cases?

harshilsharma63:
Also, @abhijit-singh since we agree this is a rare edge case, lets queue it up for v2 instead of spending time on this right now

hmhealey (approved):
This looks good to me.

I'm wondering a bit if it might make sense to store a unique ID for the "next" survey in the config to avoid having to deeply compare the two surveys in this way and to let people have identical surveys, but I can't think of any significant reason that that approach could even be better

harshilsharma63:
> This looks good to me.
> 
> I'm wondering a bit if it might make sense to store a unique ID for the "next" survey in the config to avoid having to deeply compare the two surveys in this way and to let people have identical surveys, but I can't think of any significant reason that that approach could even be better

Yeah this isn't a bottleneck. We can do it later if need arises.

yasserfaraazkhan (approved):
🎉

PR #27, opened 2024-05-23
Updated post component to handle prematurely/manually ended survey

harshilsharma63:
#### Summary
Updated survey post component to add logic to check for prematurely/manually ended survey to correctly show up as expired survey.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-58362


server/app/survey_response.go
```diff
@@ -323,3 +334,56 @@ func (a *UserSurveyApp) hasRatingGroupChanged(oldRating, newRating int) bool {
...
+	// verify post belongs to the user's DM
+	channel, appErr := a.api.GetChannel(post.ChannelId)
+	if appErr != nil {
+		a.api.LogError("HandleRefreshSurveyPost: failed to channel by ID", "channelID", post.ChannelId, "error", appErr.Error())
+		return errors.Wrapf(errors.New(appErr.Error()), "HandleRefreshSurveyPost: failed to get channel by ID, channelID: %s", post.ChannelId)
+	}
+
+	if !strings.Contains(channel.Name, userID) {
+		return fmt.Errorf("HandleRefreshSurveyPost: post does not belong to the user, postID: %s, userID: %s", postID, userID)
+	}
```
larkox, line 358: We check the channel contains the user, but we don't check the post is created by the bot, do we? There is probably no real way to exploit such vulnerability, so 0/5 on whether to add the check or not.

Related, but even more difficult to exploit, you are checking the channel is with the user, but not necessarily a DM with the bot.

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -43,6 +43,12 @@ function SurveyPost({post}: CustomPostTypeComponentProps) {
         }
     }, [responses]);
 
+    useEffect(() => {
+        if (!surveyExpired) {
+            client.refreshSurveyPost(post.id);
+        }
+    }, [post.id, surveyExpired]);
```
larkox, line 50: The post id should never change, and we only expect the survey to move from not expired to expired, so... I would say this can have an empty dependency list (only runs on mount). But 0/5.

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -224,7 +230,17 @@ function SurveyPost({post}: CustomPostTypeComponentProps) {
...
                     <div className='surveyMessage submitted'>
-                        {`Survey expired on ${surveyExpireAtDate?.toLocaleDateString()}.`}
+                        {
+                            surveyExpireAtDate &&
+                            (`Survey expired on ${surveyExpireAtDate?.toLocaleDateString()}.`)
+                        }
+
+                        {
+
+                            /*This is for the rare case of somehow the survey completely missing from system, so we might not its expiry*/
```
larkox, line 240:
```suggestion
                            /*This is for the rare case of somehow the survey completely missing from system, so we might not know its expiry*/
```

larkox (approved):
Looks good. Only a couple of nits.

server/app/survey_response.go
harshilsharma63: I should mention this in code, but this is intentional. Lets say the post was created by the bot, and the bot's ID was, say, `bot_id_1`. If, now, for any reason the bot is deleted, then plugin will create another both with the same name but different ID as IDs are pseudo-random. So, checking for bot ID isn't reliable.

webapp/src/components/surveyPost/surveyPost.tsx
harshilsharma63: Done.

webapp/src/components/surveyPost/surveyPost.tsx
harshilsharma63: Done.

harshilsharma63:
It shows 44 commits because this branch was created from `ui_refresh` branch, which is now merged into `main` and GitHub still shows the old commit count.

webapp/src/components/surveyPost/surveyPost.tsx
```diff
@@ -224,7 +230,17 @@ function SurveyPost({post}: CustomPostTypeComponentProps) {
                 {
                     disabled && surveyExpired &&
                     <div className='surveyMessage submitted'>
-                        {`Survey expired on ${surveyExpireAtDate?.toLocaleDateString()}.`}
+                        {
+                            surveyExpireAtDate &&
```
hmhealey, line 234: 0/5 but it might be worth turning these two cases into a ternary so that it's clear that they're handling opposite cases. I am a bit leery about the amount of conditional logic in this JSX, but I do really like how you've laid it out with a lot of whitespace to keep it more readable

harshilsharma63: ternary operator makes so much sense. Done!

PR #39, opened 2024-06-13
Text updates

harshilsharma63:
#### Summary
Updated following-
1. Changed bot name to `UserFeedback`.
2. Updated welcome text.
3. Updated survey text displayed on mobile.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-58460

### DIFF
--- a/server/app/survey.go
+++ b/server/app/survey.go
@@ -23,6 +23,8 @@ const (
 	surveySentValue = "survey_sent"
 
 	cacheValidityUserTeamFilter = 7200 // 2 hours in seconds
+
+	surveyPostMessage = "Help us improve your experience. Use Mattermost in a web browser or the desktop app to take the survey."
 )
 
 func (a *UserSurveyApp) SaveSurvey(survey *model.Survey) error {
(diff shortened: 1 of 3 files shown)


server/app/survey.go
```diff
@@ -23,6 +23,8 @@ const (
 	surveySentValue = "survey_sent"
 
 	cacheValidityUserTeamFilter = 7200 // 2 hours in seconds
+
+	surveyPostMessage = "Help us improve your experience. Use Mattermost in a web browser or the desktop app to take the survey."
```
larkox, line 27: Do we want this to say "Mattermost" or "the app"?

larkox (approved):
Just one minor thing. Approving just to speed up things.

devinbinnie (approved):

server/app/survey.go
larkox: The problem is white labelling. I think we try to avoid any reference to Mattermost in this plugin because of that.

harshilsharma63: This point was already discussed and this is the final text after all that discussion in the User Survey channel.

PR #47, opened 2024-06-27
Update survey date-time picker to select in user's local time zone

harshilsharma63:
#### Summary
Update survey date-time picker to select in user's local time zone.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-58497

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -54,6 +54,7 @@
     "classnames": "2.5.1",
     "core-js": "3.36.0",
     "date-fns": "3.3.1",
+    "date-fns-tz": "3.1.3",
     "enzyme": "3.11.0",
     "enzyme-to-json": "3.6.2",
     "mattermost-redux": "5.33.1",
--- a/webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
+++ b/webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
@@ -28,18 +28,23 @@ function SurveyDateTime({id, setSaveNeeded, onChange, config, setInitialSetting}
     }, [config.PluginSettings.Plugins, id, setInitialSetting]);
 
     const helpText = useMemo(() => {
-        let line1 = '';
+        let line1;
 
         if (surveyDate) {
             const localDateString = format(surveyDate, "MMMM d, yyy 'at' HH:mm O");
             const utcDateString = formatInTimeZone(surveyDate, 'UTC', "MMMM d, yyy 'at' HH:mm 'UTC'");
 
-            line1 = `Scheduled for ${localDateString} (${utcDateString})\n\n`;
+            line1 = (<p>{`Scheduled for ${localDateString} (${utcDateString})\n\n`}</p>);
         }
 
-        const line2 = 'A bot message containing the survey will start being sent to all users at the selected date and time. Delivery will occur gradually, so the exact time may vary.';
+        const line2 = (<p>{'A bot message containing the survey will start being sent to all users at the selected date and time. Delivery will occur gradually, so the exact time may vary.'}</p>);
 
-        return line1 + line2;
+        return (
+            <React.Fragment>
+                {line1}
+                {line2}
+            </React.Fragment>
+        );
     }, [surveyDate]);
 
     const saveSettings = useCallback((setting: DateTimeConfig) => {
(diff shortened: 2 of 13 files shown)


webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -1,86 +1,61 @@
...
 import SurveyDateSelector from 'components/surveyDateSelector/surveyDateSelector';
 import SurveyTimeSelector from 'components/surveyTimeSelector/surveyTimeSelector';
 import type {CustomSettingChildComponentProp} from 'components/systemConsole/index';
 
-import type {DateTimeConfig} from 'types/plugin';
-
 import './style.scss';
-
-const DEFAULT_SURVEY_TIME = '00:00';
+import type {DateTimeConfig} from 'types/plugin';
```
larkox, line 14: Surprised this is not failing on CI. This import should go before the unnamed style import.

larkox (approved):

esarafianou (commented):
I'm good with the new dependency for the timezone. 

I didn't review the rest of the changes as skimming through them, they are not related to security controls

webapp/package.json
```diff
@@ -54,6 +54,7 @@
     "classnames": "2.5.1",
     "core-js": "3.36.0",
     "date-fns": "3.3.1",
+    "date-fns-tz": "3.1.3",
```
hmhealey, line 57: The web app currently uses `moment-timezone` for this, but given Moment is legacy at this point, this seems like a better bet

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -1,86 +1,61 @@
...
 
 function SurveyDateTime({id, setSaveNeeded, onChange, config, setInitialSetting}: CustomSettingChildComponentProp) {
-    const [surveyTime, setSurveyTime] = useState<string>(DEFAULT_SURVEY_TIME);
     const [surveyDate, setSurveyDate] = useState<Date>();
 
-    const [surveyDateTime, setSurveyDateTime] = useState<Date>();
+    // sets the date picker and time picker to the values saved in config on load
+    useEffect(() => {
```
hmhealey, line 20: Doing this in a `useEffect` looks a bit odd since it's something you want run on first load, but I can't think of any real reason we shouldn't do that since this isn't a performance-sensitive area

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
```diff
@@ -1,86 +1,61 @@
...
-            time: DEFAULT_SURVEY_TIME,
-        };
+        if (surveyDate) {
+            const localDateString = format(surveyDate, "MMMM d, yyy 'at' HH:mm O");
+            const utcDateString = formatInTimeZone(surveyDate, 'UTC', "MMMM d, yyy 'at' HH:mm 'UTC'");
 
-        if (dateTimeConfig?.time) {
-            setSurveyTime(dateTimeConfig.time);
-            initialConfig.time = dateTimeConfig.time;
+            line1 = `Scheduled for ${localDateString} (${utcDateString})\n\n`;
```
hmhealey, line 37: 0/5 since this is an issue with the existing code, but I think these lines should technically be two different `<p>` tags rather than one tag with newlines in it

webapp/src/utils/utils.tsx
```diff
@@ -11,6 +11,16 @@ function uuid() {
     return uuid.substring(uuid.lastIndexOf('/') + 1);
 }
 
+function dateToUnixTimestamp(date: Date): number {
+    return Math.floor(date.getTime());
```
hmhealey, line 15: Why do a floor here? I would've thought this would always return an integer because they're Unix timestamps

hmhealey (approved):
A few questions and comments, but nothing that's blocking

webapp/package.json
harshilsharma63: yeah, date fns tz works really well with date fns and the whole idea is to work with native Date objects, not a custom ones like Moment does.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: the config object received by plugin is empty on first load, so we need to keep reacting to its change for a while.

webapp/src/utils/utils.tsx
harshilsharma63: Its not needed now. It was needed previously when I was storing seconds, not milliseconds. Removed now.

webapp/src/components/systemConsole/surveyDateTime/surveyDateTime.tsx
harshilsharma63: Done.

abhijit-singh (approved):

PR #48, opened 2024-06-27
Metadata in report

harshilsharma63:
#### Summary
Added server metadata to report.

Sample metadata YAML content-

```yaml
version: 1
type: plugin-support-package
generatedat: 1719483621782
serverversion: 9.10.0
serverid: iqnr185cttr8xnxjggdodr3mia
licenseid: ""
customerid: ""
extras:
  plugin_id: com.mattermost.user-survey
  plugin_version: 1.0.0

```

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-58688

Depends on https://github.com/mattermost/mattermost/pull/27490

### DIFF
--- a/server/api/utils.go
+++ b/server/api/utils.go
@@ -37,7 +37,7 @@ var MediaContentTypes = [...]string{
 
 // WriteFileResponse is copied from Mattermost server https://github.com/mattermost/mattermost/blob/f0121d4f23e065a05e63fd452486570271c6a28f/server/platform/shared/web/files.go#L38
 func WriteFileResponse(filename string, contentType string, contentSize int64, lastModification time.Time, webserverMode string, fileReader io.ReadSeeker, forceDownload bool, w http.ResponseWriter, r *http.Request) {
-	w.Header().Set("Cache-Control", "private, max-age=86400")
+	w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 
 	if contentSize > 0 {
--- a/server/app/app.go
+++ b/server/app/app.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	mmModel "github.com/mattermost/mattermost/server/public/model"
 	"github.com/mattermost/mattermost/server/public/plugin"
 	"github.com/mattermost/mattermost/server/public/pluginapi"
 
@@ -15,15 +16,24 @@ type UserSurveyApp struct {
 	apiClient  *pluginapi.Client
 	botID      string
 	debugBuild bool
+	manifest   *mmModel.Manifest
 }
 
-func New(api plugin.API, store store.Store, getConfigFunc func() *model.Config, driver plugin.Driver, debugBuild bool) (*UserSurveyApp, error) {
+func New(
+	api plugin.API,
+	store store.Store,
+	getConfigFunc func() *model.Config,
+	driver plugin.Driver,
+	debugBuild bool,
+	manifest *mmModel.Manifest,
+) (*UserSurveyApp, error) {
 	app := &UserSurveyApp{
 		api:        api,
 		store:      store,
 		getConfig:  getConfigFunc,
 		apiClient:  pluginapi.NewClient(api, driver),
 		debugBuild: debugBuild,
+		manifest:   manifest,
 	}
 
 	err := app.ensureSurveyBot()
--- a/server/app/app.go
+++ b/server/app/app.go
@@ -1,7 +1,6 @@
 package app
 
 import (
-	mmModel "github.com/mattermost/mattermost/server/public/model"
 	"github.com/mattermost/mattermost/server/public/plugin"
 	"github.com/mattermost/mattermost/server/public/pluginapi"
 
@@ -16,7 +15,6 @@ type UserSurveyApp struct {
 	apiClient  *pluginapi.Client
 	botID      string
 	debugBuild bool
-	manifest   *mmModel.Manifest
 }
 
 func New(
@@ -25,7 +23,6 @@ func New(
 	getConfigFunc func() *model.Config,
 	driver plugin.Driver,
 	debugBuild bool,
-	manifest *mmModel.Manifest,
 ) (*UserSurveyApp, error) {
 	app := &UserSurveyApp{
 		api:        api,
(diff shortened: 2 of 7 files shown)


harshilsharma63:
As soon as https://github.com/mattermost/mattermost/pull/27465 is merged, we'll create a new release for the `public` package and update the dependency in this plugin.

CI job will fail meanwhile.

harshilsharma63:
@catalintomai is there anything else you want to include in the metadata regarding the survey? I've added the NPS score so far.

catalintomai:
> @catalintomai is there anything else you want to include in the metadata regarding the survey? I've added the NPS score so far.

@harshilsharma63 - I wouldn't include the NPS score as part of the metadata section - imo it should be part of a separate metrics module. For metadata module - I included references in the ticket + you can take a look at https://github.com/mattermost/mattermost/pull/27465

harshilsharma63:
> > @catalintomai is there anything else you want to include in the metadata regarding the survey? I've added the NPS score so far.
> 
> @harshilsharma63 - I wouldn't include the NPS score as part of the metadata section - imo it should be part of a separate metrics module. For metadata module - we included references in the ticket + you can take a look at [mattermost/mattermost#27465](https://github.com/mattermost/mattermost/pull/27465)

Okay, will remove it from server metadata. Its already included in a separate file in the downloaded zip report.

ifoukarakis:
> @catalintomai is there anything else you want to include in the metadata regarding the survey? I've added the NPS score so far.

Perhaps the plugin's ID and version can also be added. This way we can have a better understanding on what to expect in the `extras` field.

harshilsharma63:
> > @catalintomai is there anything else you want to include in the metadata regarding the survey? I've added the NPS score so far.
> 
> Perhaps the plugin's ID and version can also be added. This way we can have a better understanding on what to expect in the `extras` field.

@ifoukarakis those are already added by default, check the sample YAML in the original post.

server/app/report.go
```diff
@@ -224,3 +231,42 @@ func (a *UserSurveyApp) generateSurveyMetadataFile(survey *model.Survey, key str
...
+	if err != nil {
+		a.api.LogError("generateServerMetadataFile: failed to encode server metadata via YAML encoder", "surveyID", survey.ID, "key", key, "filePath", filePath, "error", err.Error())
+		return "", errors.Wrapf(err, "generateServerMetadataFile: failed to encode server metadata via YAML encoder, surveyID: %s, key: %s, filePath: %s", survey.ID, key, filePath)
+	}
+
+	return filePath, nil
+}
+
+func (a *UserSurveyApp) generateServerMetadata() (*mmModel.Metadata, error) {
+	manifest := a.manifest
```
catalintomai, line 261: @harshilsharma63 , as mentioned in another comment, it'd be good to align to Ibrahim's metadata package -see Metadata structure in https://github.com/mattermost/mattermost/pull/27465/files.

hanzei: https://github.com/mattermost/mattermost/pull/27490 will ensure that all plugins use the same format.

harshilsharma63: @catalintomai I didn't understand your comment, but I'm guessing its fixed after using @hanzei 's changes?

server/app/report.go
```diff
@@ -224,3 +228,15 @@ func (a *UserSurveyApp) generateSurveyMetadataFile(survey *model.Survey, key str
 
 	return filePath, nil
 }
+
+func (a *UserSurveyApp) generateServerMetadataFile(key string) (string, error) {
+	fileDir := path.Join(os.TempDir(), "survey_report", key)
```
catalintomai, line 233: @harshilsharma63 - is "survey_report" the best name for the content of the metadata file? It is not technically a report, right?

harshilsharma63: @catalintomai @catalintomai this not the file name, its the dir name where I store all survey reports - `/tmp/survey_report/<random_key>/survey_report.zip`. The metadata file name isn't configurable, its hardcoded to be `metadata.yaml` in Ben's PR.

catalintomai (approved):

server/api/utils.go
```diff
@@ -37,7 +37,7 @@ var MediaContentTypes = [...]string{
 
 // WriteFileResponse is copied from Mattermost server https://github.com/mattermost/mattermost/blob/f0121d4f23e065a05e63fd452486570271c6a28f/server/platform/shared/web/files.go#L38
 func WriteFileResponse(filename string, contentType string, contentSize int64, lastModification time.Time, webserverMode string, fileReader io.ReadSeeker, forceDownload bool, w http.ResponseWriter, r *http.Request) {
-	w.Header().Set("Cache-Control", "private, max-age=86400")
+	w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
```
larkox, line 40: For my own knowledge, do you mind explaining this change?

server/app/app.go
```diff
@@ -15,15 +16,24 @@ type UserSurveyApp struct {
...
+	debugBuild bool,
+	manifest *mmModel.Manifest,
+) (*UserSurveyApp, error) {
 	app := &UserSurveyApp{
 		api:        api,
 		store:      store,
 		getConfig:  getConfigFunc,
 		apiClient:  pluginapi.NewClient(api, driver),
 		debugBuild: debugBuild,
+		manifest:   manifest,
```
larkox, line 36: I fail to see where the manifest is used in all this. Why is it needed?

larkox (changes requested):
Maybe I am blind, but marking as changes requested since the manifest related change seems not needed.

server/api/utils.go
harshilsharma63: Previously, I was allowing clients to cache the file. As we can export a report from an in-progress survey, lets say I generated a report when there were 10 responses. Now there are 15 responses. If I generate the report again, the browser will return the cached file instead of fetching a fresh file. So, I disabled caching for files to force clients to always re-generate a report to avoid stale data.

server/app/app.go
harshilsharma63: Nice catch! I was using it before, but then wasn't needed. I've removed it now.

hanzei (changes requested):
CI is not happy. I think you have to run `go mod tidy`.

larkox (approved):

hanzei (approved):

PR #51, opened 2024-07-18
Added new team filter customisation options

harshilsharma63:
#### Summary
Added more control to how team filters work.
Screenshot 2024-07-18 at 4 14 32 PM

#### Ticket Link
Fixes - https://mattermost.atlassian.net/browse/MM-59564

### DIFF
--- a/server/app/job.go
+++ b/server/app/job.go
@@ -82,7 +82,8 @@ func (a *UserSurveyApp) startNewSurveyIfNeeded() error {
 
 		surveyFromConfig := &model.Survey{
 			ID:              utils.NewID(),
-			ExcludedTeamIDs: config.TeamFilter.FilteredTeamIDs,
+			FilterTeamIDs:   config.TeamFilter.FilteredTeamIDs,
+			TeamFilterType:  config.TeamFilter.FilterType,
 			CreateAt:        now,
 			UpdateAt:        now,
 			StartTime:       startTime.UnixMilli(),
--- a/server/app/survey.go
+++ b/server/app/survey.go
@@ -121,6 +121,10 @@ func (a *UserSurveyApp) ShouldSendSurvey(userID string, survey *model.Survey) (b
 	return !inExcludedTeam, nil
 }
 
+//func (a *UserSurveyApp) userPassesSurveyTeamFilter(userID string, survey *model.Survey) (bool, error) {
+//	if survey.team
+//}
+
 func (a *UserSurveyApp) GetSurveyPostIDSentToUser(userID, surveyID string) (string, error) {
 	postID, appErr := a.api.KVGet(utils.KeyUserSurveySentStatus(userID, surveyID))
 	if appErr != nil {
--- a/server/app/survey.go
+++ b/server/app/survey.go
@@ -221,7 +221,7 @@ func (a *UserSurveyApp) ensureSurveyBot() error {
 func (a *UserSurveyApp) userInExcludedTeams(userID string, survey *model.Survey) (bool, error) {
 	// no need to check for memberships if
 	// no teams were excluded from the survey
-	if len(survey.ExcludedTeamIDs) == 0 {
+	if len(survey.FilterTeamIDs) == 0 {
 		return false, nil
 	}
 
@@ -244,7 +244,7 @@ func (a *UserSurveyApp) userInExcludedTeams(userID string, survey *model.Survey)
 	}
 
 	filteredTeamMap := map[string]bool{}
-	for _, teamID := range survey.ExcludedTeamIDs {
+	for _, teamID := range survey.FilterTeamIDs {
 		filteredTeamMap[teamID] = true
 	}
 
--- a/server/app/survey.go
+++ b/server/app/survey.go
@@ -113,12 +113,12 @@ func (a *UserSurveyApp) ShouldSendSurvey(userID string, survey *model.Survey) (b
 		return false, nil
 	}
 
-	userPassesTeamFIlter, err := a.userPassesSurveyTeamFilter(userID, survey)
+	userPassesTeamFilter, err := a.userPassesSurveyTeamFilter(userID, survey)
 	if err != nil {
 		return false, errors.Wrap(err, "ShouldSendSurvey: failed to check if user passes team filter or not")
 	}
 
-	return userPassesTeamFIlter, nil
+	return userPassesTeamFilter, nil
 }
 
 func (a *UserSurveyApp) userPassesSurveyTeamFilter(userID string, survey *model.Survey) (bool, error) {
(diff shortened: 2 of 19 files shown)


abhijit-singh:
Thanks @harshilsharma63. Can you put this on a test server?

server/app/survey_test.go
```diff
@@ -157,9 +159,104 @@ func TestShouldSendSurvey(t *testing.T) {
...
+			Status:         "in_progress",
+			FilterTeamIDs:  []string{"team_id_1"},
+			TeamFilterType: model.TeamFilterExcludeSelected,
+		}
+		should, err := th.App.ShouldSendSurvey("user_id", survey)
+		require.NoError(t, err)
+		require.False(t, should)
+	})
+
+	t.Run("should send as user is include team", func(t *testing.T) {
```
marianunez, line 172: Should we have a test that it doesn't send if it's in excluded teams? Seems that the tests below are when the teams are empty

marianunez (approved):

server/app/survey_test.go
harshilsharma63: Because the existing behaviour was to not send survey to user in specified teams, the tests for those already exists, for example we have this existing test - `should not send as user is in a excluded team`

abhijit-singh (approved):
Looks perfect functionally @harshilsharma63. 

A couple of small things: 
1. The radio button and label seem to be sticking to each other right now. There should be 8px of space between them. 
image

2. There is a slightly weird jumping behaviour when switching from `Send to everyone` and any of the other options. I was expecting everything below it to move up instead of redistributing the space between the fields. But I think that's difficult because we need to keep the total container height constant for the collapsing functionality so it is okay to keep this as is. 
https://github.com/user-attachments/assets/43d1101c-a91d-4544-be55-fc58e641e980

server/app/survey_test.go
marianunez: Ah ok, missed those in the diff but I see them now when looking at the whole fle. Thanks for clarifying 👍

harshilsharma63:
@abhijit-singh I've fixed those those UI now and have updated the plugin on test server.

abhijit-singh:
Thanks @harshilsharma63! Looks good! Just one more tiny thing that I noticed today — the field label is not aligned with the first radio button. We can add a margin of 9px above the input options to align them.

image

harshilsharma63:
@abhijit-singh done.

server/app/job.go
```diff
@@ -82,7 +82,8 @@ func (a *UserSurveyApp) startNewSurveyIfNeeded() error {
 
 		surveyFromConfig := &model.Survey{
 			ID:              utils.NewID(),
-			ExcludedTeamIDs: config.TeamFilter.FilteredTeamIDs,
+			FilterTeamIDs:   config.TeamFilter.FilteredTeamIDs,
+			TeamFilterType:  config.TeamFilter.FilterType,
```
hmhealey, line 86: No changes to make here, but I wonder if this might've been easier to implement if this was represented internally as two arrays: `IncludedTeamIDs` and `ExcludedTeamIDs` since that would've been my first thought of how to store this. That seems like it might've made the code checking those arrays simpler, but it does mean that it's possible for this configuration to end up in a way that the UI can't represent

server/app/survey.go
```diff
@@ -214,10 +237,10 @@ func (a *UserSurveyApp) ensureSurveyBot() error {
 	return nil
 }
 
-func (a *UserSurveyApp) userInExcludedTeams(userID string, survey *model.Survey) (bool, error) {
+func (a *UserSurveyApp) userMemberOfFilteredTeams(userID string, survey *model.Survey) (bool, error) {
```
hmhealey, line 240: Do we need to do anything to clear the cache for this method if the admin changes the new option to include or exclude users from a given team?

server/store/migrations/000014_survey_table_add_team_filter_type_column.up.sql
```diff
@@ -0,0 +1,7 @@
+{{- /* you annot add a new colum with not null constraint without a default value */ -}}
+{{- /* so, we're adding the column and setting the default value we need to set in existing rows */ -}}
+{{- /* then removing the default value from column as its no longer needed */ -}}
```
hmhealey, line 3: This style of comment writing is a bit odd to me. Could we not just write this as a regular SQL comment like this? Also, there's a couple letters missing in the first line
```suggestion
-- you cannot add a new column with not null constraint without a default value
-- so, we're adding the column and setting the default value we need to set in existing rows
-- then removing the default value from column as its no longer needed
```

hmhealey (approved):

server/app/job.go
harshilsharma63: The list teams never change on changing the options, the filter type just defines how the list of teams is to be treated. As such, its of no benefits to store them as two lists.

At a time, there can only be one behaviour - you include the list, or exclude, not both,. So no use of storing two lists.

server/app/survey.go
harshilsharma63: The cache is per user, per survey (cache key uses survey ID and user ID), so, existing cache, or even cache from one survey, won't affect another survey. This also works fine if there are two surveys ending and starting within the span of cache expiry duration.

server/store/migrations/000014_survey_table_add_team_filter_type_column.up.sql
harshilsharma63: Done.

server/app/survey.go
```diff
@@ -113,12 +113,35 @@ func (a *UserSurveyApp) ShouldSendSurvey(userID string, survey *model.Survey) (b
 		return false, nil
 	}
 
-	inExcludedTeam, err := a.userInExcludedTeams(userID, survey)
+	userPassesTeamFIlter, err := a.userPassesSurveyTeamFilter(userID, survey)
```
esarafianou, line 116: nit
```suggestion
	userPassesTeamFilter, err := a.userPassesSurveyTeamFilter(userID, survey)
```

server/app/survey.go
```diff
@@ -214,10 +237,10 @@ func (a *UserSurveyApp) ensureSurveyBot() error {
 	return nil
 }
 
-func (a *UserSurveyApp) userInExcludedTeams(userID string, survey *model.Survey) (bool, error) {
+func (a *UserSurveyApp) userMemberOfFilteredTeams(userID string, survey *model.Survey) (bool, error) {
 	// no need to check for memberships if
 	// no teams were excluded from the survey
```
esarafianou, line 242: nit
```suggestion
	// no teams were filtered from the survey
```

esarafianou (approved):

server/app/survey.go
harshilsharma63: Done.

server/app/survey.go
harshilsharma63: Done.

PR #53, opened 2024-07-23
Removed redundant mutex from migration and used MM migration timeout setting for plugin migration timeout

harshilsharma63:
#### Summary
Applied following changes to migration code-
1. Removed manually managed mutex as morph already manages itself via the `morph.WithLock()` function which we're already using.
2. Removed the hardcoded migration timeout config and used Mattermost's timeout config.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-59845

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -110,13 +110,14 @@ func (p *Plugin) createStoreParams() (*store.Params, error) {
 	}
 
 	return &store.Params{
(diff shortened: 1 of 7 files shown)


larkox (approved):

esarafianou:
@harshilsharma63 I skimmed through the code and there are no security related changes. No need from a proper review from us.

hmhealey (approved):

